PR #15, opened 2015-11-10
Chartjs v2 integration

cesar-tonnoir:


cesar-tonnoir:
@xaun @alexnoox can you review please?

src/components/chart/chart.directive.coffee
```diff
@@ -1,73 +1,40 @@
...
-            angular.extend(options,data.options)
+        scope.draw = (chartData) ->
+
+          angular.merge(chartData.options, {
+            scales: {
+              xAxes: [{
+                display: false
+              }]
+            }
+          }) if chartData.options.showXLabels? && !chartData.options.showXLabels
```
alexnoox: Can you not "inline" if on multiple lines instruction? On a block like this the if may be missed.

alexnoox:
cnapproval

Issue #6, opened 2015-11-03
Accounts comparison - selected accounts list is cleared upon company selection / checkbox ticking


cesar-tonnoir:
Instead of completely clearing the accounts list, we should keep in the list the accounts that have previously been selected, and that match the new criteria.

Example:

````Cucumber
Given that I have only "CompanyA" enabled, and I have selected "Account1" and "Account2",
When I enable "CompanyB"
Then it should keep "CompanyA - Account1" and "CompanyA - Account2" in the selection list
````

Other example:

````Cucumber
Given that I have "CompanyA" and "CompanyB" selected, in "compare matching accounts" mode, with "Account1" selected
When I untick the checkbox,
Then it should keep "CompanyA - Account1" and "CompanyB - Account1" in the selection list
````

See the idea @xaun ?

cesar-tonnoir:
https://maestrano.atlassian.net/browse/IMPAC-56

PR #32, opened 2015-11-11
Fix for account comparison - single organization

xaun:
Prevents running comparison mode or loading comparison mode from metadata settings if there aren't two or more organisations available on the widget.
This catches discrepancies between widget.metadata and the potential changes in availability of multiple organisations.

@cesar-tonnoir @ouranos This merge request will fix the widget being unusable in some cases.

On a seperate branch (xaun/rework/accounts-comparison) I have removed the grouping functionality from the front-end and hooked it up to the grouping function I wrote in Impac API. It's working much better, but it is not finished, I need to write some of the front-end behaviour logic outlined here:  https://github.com/maestrano/impac-angular/issues/6

:v:

### DIFF
--- a/src/components/widgets/accounts-comparison/accounts-comparison.directive.coffee
+++ b/src/components/widgets/accounts-comparison/accounts-comparison.directive.coffee
@@ -29,7 +29,7 @@ module.controller('WidgetAccountsComparisonCtrl', ($scope, $q, ChartFormatterSvc
       value: false,
       onChangeCallback: $scope.multiCompanyComparisonOnChange
     }]
-    if angular.isDefined w.metadata.comparison_mode
+    if angular.isDefined(w.metadata.comparison_mode) && w.metadata.organization_ids.length > 1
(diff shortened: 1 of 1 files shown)


cesar-tonnoir:
@xaun @ouranos what about bundling our minor bugs fixes in another branch (like fixes/v1.0.2), which would be merged every two weeks (or before if needed)? This would prevent us from creating a new subversion every day with only 1 line of code changed?...

ouranos:
You don't have  to create a patch version every time you changes are merged. 
Bug fixes can be merged in  `1.0` (which is then regularly merged in `master`).
Once you feel a new patch version is needed, either because of a critical fix or because enough fixes are in, you can bump and tag.

If someone absolutely need the fix without targetting `master`, they can always target the `1.0` branch instead of a specific `1.0.x` version. 

Looks at this branch of devise for an example: 
https://github.com/plataformatec/devise/commits/v3.2

Rails does  the same too, their patch releases are usually pretty big:
https://github.com/rails/rails/compare/v4.2.4...v4.2.5.rc1

PR #48, opened 2015-11-24
Improved front-end logic for comparable and non-comparable accounts

xaun:
- when the returned accounts data is from only one organisation, comparison mode is disabled.
- if comparison mode selection was saved and then returned accounts data is from only one organisation, an error message is displayed and the ability to deselect comparison mode is available. Once deselection is made, the comparison mode is disabled.


xaun:
Please review @cesar-tonnoir :8ball:

cesar-tonnoir:
:+1:

PR #50, opened 2015-11-25
Safari fix: canvas isn't draw when elemt is not visible

alexnoox:


alexnoox:
@cesar-tonnoir Check and merge

cesar-tonnoir:
should be merged to 1.0 as its a bugfix

PR #52, opened 2015-11-25
Safari fix: canvas isn't draw when elemt is not visible

cesar-tonnoir:


cesar-tonnoir:
thanks @alexnoox ! :+1:

PR #63, opened 2015-12-07
Bugfix/frontend

alexnoox:


alexnoox:
@cesar-tonnoir I just added valid and cancel buttons to editable title in this PR, please review.

PR #72, opened 2015-12-09
KPI loader when attaching a KPI

alexnoox:


alexnoox:
@cesar-tonnoir To review and merge

PR #103, opened 2016-02-04
add AED and IDR to currencies list

cesar-tonnoir:


cesar-tonnoir:
> need changelog update

PR #111, opened 2016-02-17
[IMPAC-35] Better time period selection

cesar-tonnoir:
> Need specs for directives setting-time-period, setting-time-slider, setting-time-presets


cesar-tonnoir:
@xaun can you have a look when you have time please?
It's going to be merged only next week after the backend is updated and the specs finished

PR #113, opened 2016-02-19
[Example only] Net Sales widget

alachaum:


src/services/widgets/widgets.svc.coffee
```diff
@@ -222,13 +220,11 @@ angular
...
-              angular.extend widget, updatedWidget
-              deferred.resolve(widget)
+              angular.extend widget, success.data
             (error) ->
-              $log.error("ImpacWidgetsSvc: cannot update widget: #{widget.id}")
-              deferred.reject(error)
-          )
+              angular.merge(widget, opts)
+              $log.warn("ImpacWidgetsSvc: unable to remotely save widget state (id: #{widget.id})")
+          ).finally( -> deferred.resolve(widget))
```
cesar-tonnoir: Don't think we can do it this way: the consumer of the service sometimes waits for a rejection of the deferred object.
Ex: widget name edition - the name has to fallback to its previous value in case of failure

alachaum: Why? The name should still be updated. However we should trigger a retry
instead and/or alert the user that the name couldn't be saved.

Arnaud Lachaume
Chief Technology Officer
*Maestrano*

*USA*
506 2nd Ave Suite 1400 Seattle, WA 98104
*Ph| *+1 (206) 582-5075

*Australia*
Suite 504, 46 Market Street, Sydney, NSW 2000
*Ph*| +61 (0) 2 9279 4344 *Mob*| +61 (0) 449 770 094

arnaud.lachaume@maestrano.com | www.maestrano.com <http://maestrano.com/>

cesar-tonnoir: Yep, you're right, it should trigger an alert...
I'm going to change this

cesar-tonnoir:
merged with #118

PR #120, opened 2016-02-22
[IMPAC-125] Fix/gulp build

xaun:
@ouranos I have merged your gulp work - please review

@cesar-tonnoir


bower.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "main": [
     "dist/impac-angular.js",
-    "src/impac-angular.less"
+    "dist/impac-angular.less"
```
xaun: @ouranos I think this is fine, but just confirming that parent apps less overrides, etc, will still function as they were when pointing to `src/impac-angular.less`? What's different about the new concatenated dist less file that allows us to do this now?

xaun:
no idea why travis is breaking - will look at it in the morning.

ouranos:
LGTM.  Please test in an MNOE project before merging :)

xaun:
@cesar-tonnoir @ouranos tested on mnoe and good to merge :beers:

PR #142, opened 2016-03-15
fix(mno-currency): symbols & formatting

xaun:
Official symbols for 12/14 supported currencies, 'showName' filter argument changed to 'ISOmode', for switching between suffixed ISO code and prefixed currency symbol display.

@cesar-tonnoir please review


cesar-tonnoir:
All good
thanks for the spec! :beers:

PR #148, opened 2016-04-01
[IMPAC-150] Params-picker setting improvements

AterStrix:


cesar-tonnoir:
Hi @AterStrix,
Thanks for submitting your changes.
I'm going to add some in-line comments. Feel free to ask questions when needed!

Cheers

cesar-tonnoir:
Hi @AterStrix,
I've reviewed your changes and unfortunately I'm afraid that a great part of it will have to be reworked. I let you check the inline remarks I will add to it, but just before, a few points:
- Please add comments to your code, otherwise it's quite hard to keep track with what you have in mind..
- Make sure that your code is consistent with the good practices that are used in the rest of the project. In particular:
  - setting-scoped functions are not supposed to be accessed from the outside of the setting's directive
  - when you code directly in a service, keep it dynamic (for instance you should refer to a dynamic parameter `dashboard[settingKey]` rather than `dashboard.status_selection`. Same thing: we shouldn't have a specific `params-picker` condition in WidgetsSvc.updateWidgetSettings()...)

src/components/widgets-settings/params-picker/params-picker.directive.coffee
```diff
@@ -14,16 +14,26 @@ module.controller('SettingParamsPickerCtrl', ($scope) ->
       'ui-floating': true,
       tolerance: 'pointer'
     }
-    setting.isInitialized = true if _.isEmpty($scope.options)
+    $scope.applyForDashboard = false;
+    $scope.applyForAll = ->
+      setting.reach = $scope.applyForDashboard;
 
+    setting.isInitialized = true if _.isEmpty($scope.options)
```
cesar-tonnoir: at initialization:
  ```
    # setting.reach is initialized based on the metadata value, or will be defaulted to 'widget'
    setting.reach = w.metadata && w.metadata[$scope.param] && w.metadata[$scope.param].reach || 'widget'

    # applyToDashboard boolean is initialized based on setting.reach
    $scope.applyToDashboard = (setting.reach == 'dashboard')
  ```

src/components/widgets-settings/params-picker/params-picker.directive.coffee
```diff
@@ -14,16 +14,26 @@ module.controller('SettingParamsPickerCtrl', ($scope) ->
...
 
-  setting.toMetadata = ->
+  setting.toMetadata = ()->
     param = {}
     param[$scope.param] = _.compact(_.map $scope.options, (statusOption) ->
       statusOption.label if statusOption.selected
     )
+    param.reach = $scope.applyForDashboard = false;
     return param
```
cesar-tonnoir: in toMetadata():
  ```
    param = {}

    # returned element has to be changed as we want to associate the "reach" attribute to the setting, and not to the widget!
    param[$scope.param] = {
      values: _.compact(_.map $scope.options...
      reach: setting.reach
    }
  ```

src/components/widgets-settings/params-picker/params-picker.directive.coffee
```diff
@@ -14,16 +14,26 @@ module.controller('SettingParamsPickerCtrl', ($scope) ->
...
 
+    setting.isInitialized = true if _.isEmpty($scope.options)
 
-  setting.toMetadata = ->
+  setting.toMetadata = ()->
     param = {}
     param[$scope.param] = _.compact(_.map $scope.options, (statusOption) ->
       statusOption.label if statusOption.selected
     )
+    param.reach = $scope.applyForDashboard = false;
```
cesar-tonnoir: in toMetadata():
  ```
    param = {}

    # returned element has to be changed as we want to associate the "reach" attribute to the setting, and not to the widget!
    param[$scope.param] = {
      values: _.compact(_.map $scope.options...
      reach: setting.reach
    }
  ```

src/components/widgets-settings/params-picker/params-picker.directive.coffee
```diff
@@ -14,16 +14,26 @@ module.controller('SettingParamsPickerCtrl', ($scope) ->
...
     return param
 
+  setting.getOptions = ->
+    return $scope.options
+
+  setting.setOptions = (options)->
+    _.assign $scope.options, options
+
```
cesar-tonnoir: "options" attribute in the setting should not be accessed from the outside of the directive as a setting is a stateless element that is initialized based on the widgets' metadata. Instead, the standard cycle should be followed:
  1. widgets' metadata are retrieved or changed
  2. WidgetsSvc.show is retrieving the widgets' data based on these metadata
  3. widget.initContext() is called to initialize the widget
  4. setting.initialize() is called for each setting of the widget

src/components/widgets-settings/params-picker/params-picker.tmpl.html
```diff
@@ -10,4 +10,7 @@
       <input type="checkbox" ng-model="parameter.selected" />
     </span>
   </div>
-</div>
\ No newline at end of file
+  <div>
+    <label><strong>Apply to all widgets</strong> <input type="checkbox" ng-true-value="'dashboard'" ng-change="applyForAll()" ng-model="applyForDashboard"></label>
+  </div>
+</div>
```
cesar-tonnoir: - `<strong>Apply to all similar widgets</strong>` => don't forget that some widgets won't include this setting
- `ng-true-value` can be removed there.
- change wording: `applyForAll()` => `toggleReach()` ; `applyForDashboard` => `applyToDashboard`
-  upon checkbox tick/untick:
  ```
    # ng-change should point to a simple toggle which is updating setting.reach. Then, setting.reach will be used everywhere as the unique source of info.

    # this way, setting.reach will always be either 'dashboard' or 'widget'

    $scope.toggleReach = ->
      setting.reach = ($scope.applyToDashboard ? 'dashboard' : 'widget')
  ```

src/components/widgets/sales-opportunities-funnel/sales-opportunities-funnel.directive.coffee
```diff
@@ -34,13 +34,17 @@ module.controller('WidgetSalesOpportunitiesFunnelCtrl', ($scope, $q, ChartFormat
   w.initContext = ->
     if $scope.isDataFound = angular.isDefined(w.content) && !_.isEmpty(w.content.opps_per_sales_stage) && hasOneOpportunity(w.content.opps_per_sales_stage)
 
-      $scope.statusOptions = _.compact _.map w.metadata.sales_stage_selection, (status) ->
+      defaultWidget = ImpacWidgetsSvc.getDefaultWidget(w.name);
```
cesar-tonnoir: Logic of having a "default widget" and dashboard.status_selection being based on this defaultWidget is not ok: what if the widget is removed from the dashboard? The status_selection values should be stored just as it is stored in the widget's metadata:
  ```
    dashboard.status_selection = {
      values: [ 'Status1', 'Status2' ]
      reach: 'dashboard'
    }
  ```
This way, even if the first widget is deleted, any new widget would be defaulted to the stored status selection.

src/services/widgets/widgets.svc.coffee
```diff
@@ -76,20 +76,52 @@ angular
...
+      if (defaultWidget)
+        defaultWidget.metadata = newMeta
+      else
+        dashboard.status_selection.push({'name': widget.name, 'metadata': newMeta})
+
+      ImpacDashboardsSvc.update dashboard.id, {'status_selection': dashboard.status_selection}
+
+    @getDefaultWidget = (widgetName)->
+      return _.find(ImpacDashboardsSvc.getCurrentDashboard().status_selection, {'name': widgetName})
+
```
cesar-tonnoir: The overall logic of applying the setting to the dashboard + to the dashboard's widgets has to be reworked as there is a simpler way to do it that would be closer to the current behaviour. Here is the logical cycle that should be followed:
  1. In WidgetsSvc.updateWidgetSettings() function:
    - add a parameter "ignoreReach", defaulted to false, that will trigger the behaviour below when relevant 
    - in the loop that is done on the widget's settings:
      if the setting contains the property "reach" and if `setting.reach == 'dashboard'` and `ignoreReach` is false
      then call the function `Dashboard.updateSetting(dashboard, setting)` (see just below)

  2. Define `DashboardsSvc.updateSetting(DHB, SET)` function:
    - call `DashboardsSvc.update(DHB.id, setting.toMetadata())` => will save the metadata to the dashboard's model
    - loop on DHB.widgets
        - loop on each widget's settings
          - if `setting.toMetadata()` has the same key as `SET.toMetadata()` (setting passed has an argument),
          then that means that the widget has a setting of the same kind that the one we are trying to push to all the dashboard's widgets.
          - => `angular.merge(widget.metadata, SET.toMetadata())`
          - => `WidgetsSvc.initWidgetSettings(widget)` (to re-initialize the settings based on the new metadata)
          - => `WidgetsSvc.updateWidgetSettings(widget, ignoreReach=true)` (to push the new metadata to the widget's model)

  3. In the concerned widget directive, `initContext()` function:
    - options should be defaulted to the `widget.metadata` if defined, then to the `dashboard.medata` if defined
    - => Allows to:
      - keep a specific config for a widget that hasn't ticked the box
      - apply the dashboard's config for the new widgets added to the dashboard

src/components/widgets-settings/params-picker/params-picker.directive.coffee
```diff
@@ -8,20 +8,29 @@ module.controller('SettingParamsPickerCtrl', ($scope) ->
   setting = {}
   setting.key = "params-picker"
   setting.isInitialized = false
+  setting.paramName = $scope.param
```
cesar-tonnoir: `setting.paramName` is not necessary there: a call can be done to `setting.toMetadata()` to identify the param name. For example, 
  ```
    setting.toMetadata()
    -----
    =>
      {
        status_selection: {
          values: ['Status1', 'Status2'],
          reach: 'dashboard'
        }
      }
  ```
  There, `_.keys(setting.toMetadata())[0]` will return `status_selection` which is our param name

src/components/widgets/sales-cycle/sales-cycle.directive.coffee
```diff
@@ -23,15 +23,20 @@ module.controller('WidgetSalesCycleCtrl', ($scope, $q, ChartFormatterSvc, $filte
   # --------------------------------------
   w.initContext = ->
     if $scope.isDataFound = angular.isDefined(w.content) && !_.isEmpty(w.content.status_average_durations)
+
+      meta = if w.metadata.status_selection.reach == 'dashboard' then ImpacDashboardsSvc.getCurrentDashboard().status_selection else w.metadata.status_selection
```
cesar-tonnoir: guarding: `if metadata.status_selection && metadata.status_selection.reach == 'dashboard'`

cesar-tonnoir: naming: "meta" is not obvious. Maybe "status_selection" ? "st_selection" ?...

cesar-tonnoir: important: **test here is not ok** - should not be testing `widget.metadata.status_selection.reach`: what about a newly added widget that wouldn't have any metadata defined?

Please follow logic described in previous comment (should be defaulted to the widget.metadata if defined, or to the dashboard.medata if defined).

Same thing for other widgets

src/components/widgets/sales-opportunities-funnel/sales-opportunities-funnel.directive.coffee
```diff
@@ -34,13 +33,17 @@ module.controller('WidgetSalesOpportunitiesFunnelCtrl', ($scope, $q, ChartFormat
   w.initContext = ->
     if $scope.isDataFound = angular.isDefined(w.content) && !_.isEmpty(w.content.opps_per_sales_stage) && hasOneOpportunity(w.content.opps_per_sales_stage)
 
-      $scope.statusOptions = _.compact _.map w.metadata.sales_stage_selection, (status) ->
```
cesar-tonnoir: **warning**, name of the param here is "sales_stage_selection" and not "status_selection": changing this would simply break the widget as the backend is not designed to receive a "status_selection" metadata parameter...

src/services/widgets/widgets.svc.coffee
```diff
@@ -74,22 +74,39 @@ angular
...
-      
-      return _self.update(widget, { metadata: meta }).then(
+
+      _self.update(widget, { metadata: meta }).then(
         (updatedWidget) ->
           if needContentReload
             _self.show(updatedWidget).finally( -> updatedWidget.isLoading = false )
       )
 
+    @updateAllSameWidgets = (widget, dashboard, settings) ->
```
cesar-tonnoir: `WidgetsSvc.updateAllSameWidgets`:
  - this logic should be placed in `DashboardsSvc`, as it concerns the update of a dashboard
  - **important**: you cannot test the widget's name to identify the widgets that have to be changed: the widget's name can be changed by the user... Please stick to the logic described in previous comment: (apply `toMetadata()` on all the settings of all the widgets ; the widgets to be updated will be the ones containing at least one setting which has the same key as the setting passed in parameter)

AterStrix: 1) I can't do that, because 2 services, according to angular, can't be in dependency of each other in the same time, so, I can't use WidgetsSvc methods from DashboardSvc directly. 
All I can is set method updateAllSameWidgets as a callback, so it will be called from DashbordSvc after dashboard updated with updateSetting method.
2) So...then widgets required them own key parameter. Because if I rewrite metadata by setting key I rewrite it in all widgets where the setting with same key is defined, even if the widget uses another list of parameters. In other words, I will rewrite settings of "sales-cycle" when saves "sales-opportunities-funnel", for example.

src/components/widgets/sales-opportunities-funnel/sales-opportunities-funnel.directive.coffee
AterStrix: I use this parameter as the one which I can update and save on dashboard. I can use appropriate parameter names but they will work only when backend starts saving them.

cesar-tonnoir: Yes, by all means. Let's keep `sales_stage_selection` for now: the backend update will be deployed next week.

src/services/widgets/widgets.svc.coffee
cesar-tonnoir: My bad for 1.: I didn't see the recursive inclusion issue. Let's just add a comment `TODO: move logic in ImpacDashboardsSvc`, I'll try to figure out a way to move it later.

For 2.: I get your point, but actually, "rewrite it in all widgets were the setting key is defined" is basically what we are trying to do here: rewrite all the widgets were the user defined a list of lead statuses (for leads), or of sales stage statuses (for opportunities). So, that's why the setting key has to be specific to the information held (regarding the other comment).

AterStrix: For 2. Looks like we have little missunderstanding if I use key of setting as a parameter for widgets filtering we recieve all widgets with the setting. For example: setting with key "param-picker" when saving "sales-cycle" will rewrite setting "param-picker" on widget(s) "opportunities funnel" too, so we write "meta.values" from "sales-cycle" to "opportunities funnel" and break the second one. I hope we understand each other now?

src/services/widgets/widgets.svc.coffee
```diff
@@ -74,22 +74,52 @@ angular
...
       )
 
+    # TODO: move logic in ImpacDashboardsSvc
+    # Sets setting for all widgets with same name
+    @updateAllSameWidgets = (widget, dashboard, settings) ->
+      # Find name of parameter
+      paramName = _.keys(settings.toMetadata())[0];
+
+      # Find widgets with same name
+      sameWidgets = _.filter dashboard.widgets, (wgt)-> wgt.name == widget.name
```
cesar-tonnoir: As explained before: you cannot test the widget's name to identify the widgets that have to be changed, because the widget's name can be changed by the user.

Try something like that:

```coffeescript
@updateAllSameWidgets = (dashboard, setting) ->
      # Find name of parameter
      paramName = _.keys(setting.toMetadata())[0];

      # Write new settings metadata to current dashboard
      ImpacDashboardsSvc.update(dashboard.id, setting.toMetadata()).then (updatedDashboard)->

        # Update new setting on current dashboard
        # TODO: can be moved to DashboardsSvc.update?
        dashboard[paramName] = updatedDashboard[paramName]

        # Set new setting to the same widgets and update them
        _.each dashboard.widgets, (wgt) ->
          # Retrieve the name of parameters attached to the widget
          # TODO: export to a helper function in WidgetsSvc
          wgtSettingsKeys = _.uniq( _.map( wgt.settings, (st) ->
            _.keys(st.toMetadata())[0]
          ))

          # The widget's metadata are updated only if the correct setting is attached to the widget
          if _.includes(wgtSettingsKeys, paramName)
            wgt.metadata[paramName] = _.cloneDeep setting.toMetadata()[paramName]
            wgt.isLoading = true
            _self.update(wgt, { metadata: wgt.metadata }).then(
              (updatedWidget) ->
                _self.show(updatedWidget).finally( -> updatedWidget.isLoading = false )
            )
```

src/components/widgets/sales-opportunities-funnel/sales-opportunities-funnel.tmpl.html
```diff
@@ -4,7 +4,7 @@
     <h4>Widget settings</h4>
 
     <div setting-organizations parent-widget="widget" class="part" deferred="::orgDeferred" />
-    <div setting-params-picker options="statusOptions" param="sales_stage_selection" parent-widget="widget" class="part" deferred="::paramsPickerDeferred" />
+    <div setting-params-picker options="statusOptions" has-reach="hasReach" param="opps_per_sales_stage" parent-widget="widget" class="part" deferred="::paramsPickerDeferred" />
```
cesar-tonnoir: Again, breaks the widget... `param="sales_stage_selection"`

src/components/widgets/sales-cycle/sales-cycle.tmpl.html
```diff
@@ -4,7 +4,7 @@
     <h4>Widget settings</h4>
 
     <div setting-organizations parent-widget="widget" class="part" deferred="::orgDeferred" />
-    <div setting-params-picker options="statusOptions" param="status_selection" parent-widget="widget" class="part" deferred="::paramsPickerDeferred" />
+    <div setting-params-picker options="statusOptions" param="status_average_durations" has-reach="hasReach" parent-widget="widget" class="part" deferred="::paramsPickerDeferred" />
```
cesar-tonnoir: Why changing `param="status_selection"` here? Have you been able to test this widget?

src/components/widgets/sales-leads-funnel/sales-leads-funnel.tmpl.html
```diff
@@ -4,7 +4,7 @@
     <h4>Widget settings</h4>
 
     <div setting-organizations parent-widget="widget" class="part" deferred="::orgDeferred" />
-    <div setting-params-picker options="statusOptions" param="status_selection" parent-widget="widget" class="part" deferred="::paramsPickerDeferred" />
+    <div setting-params-picker options="statusOptions" param="leads_per_status" has-reach="hasReach" parent-widget="widget" class="part" deferred="::paramsPickerDeferred" />
```
cesar-tonnoir: Same thing here, I don't see why the parameter naming is changed...
A metadata is an information about the widget's configuration that will be sent to the back-end. If you change its naming it can potentially break the back-end as it won't receive the parameter it's expecting.

Plus, what about the widgets that are already in production and use the original configuration param name?

src/components/widgets/sales-opportunities-funnel/sales-opportunities-funnel.tmpl.html
AterStrix: Now parameter name is the same as its name in back-end.

src/components/widgets/sales-cycle/sales-cycle.tmpl.html
AterStrix: I tested with status_selection. Then I changed the parameter name to status_average_durations, because we receive this parameter from back-end with this name. So it will be needed for correct work when 
back-end will be updated.

src/components/widgets/sales-leads-funnel/sales-leads-funnel.tmpl.html
AterStrix: I receive the parameter with the name from backend, so, if I understand correctly, I have to send it back with the same name, right?

src/services/widgets/widgets.svc.coffee
AterStrix: "paramName" in our situation is not the name like "sales_stage_selection" or "status_selection", it is key of "params-picker" setting and metadata of the setting is 
{value: ["Hot", "Cold", "etc.."], reach: "dashboard"}. But we have "params-picker" in at least three widgets, "sales-cycle", "sales-leads" and "sales-opportunities", so when I find all widgets by "params-picker" parameter name I receive all of them, and write "{value: ["Hot", "Cold", "etc.."], reach: "dashboard"}" to all of them, but "sales-opportunities" and "sales-cycle", for example, have different lists of values, so I overwrite those widgets with incorrect values. 
I will try to add extra key for all these widgets and I think it will help to find only required widgets.

src/components/widgets/sales-cycle/sales-cycle.directive.coffee
```diff
@@ -23,15 +23,23 @@ module.controller('WidgetSalesCycleCtrl', ($scope, $q, ChartFormatterSvc, $filte
   # --------------------------------------
   w.initContext = ->
     if $scope.isDataFound = angular.isDefined(w.content) && !_.isEmpty(w.content.status_average_durations)
+
+      # Takes metadata from dashboard or from widget according to 'reach' parameter
+      status_average_durations = if w.metadata.status_average_durations && w.metadata.status_average_durations.reach == 'dashboard' then ImpacDashboardsSvc.getCurrentDashboard().status_average_durations else w.metadata.status_average_durations
```
cesar-tonnoir: Hi Artem,
There is still a reference to status_average_durations here / we have to look at w.metadata.status_selection

src/components/widgets/sales-cycle/sales-cycle.tmpl.html
cesar-tonnoir: As discussed, please change it back to status_selection

src/components/widgets/sales-leads-funnel/sales-leads-funnel.directive.coffee
```diff
@@ -34,13 +34,20 @@ module.controller('WidgetSalesLeadsFunnelCtrl', ($scope, $q, ChartFormatterSvc,
   w.initContext = ->
     if $scope.isDataFound = angular.isDefined(w.content) && !_.isEmpty(w.content.leads_per_status) && hasOneLead(w.content.leads_per_status)
 
-      $scope.statusOptions = _.compact _.map w.metadata.status_selection, (status) ->
-        {label: status, selected: true} if angular.isDefined(w.content.leads_per_status[status])
+      # Takes metadata from dashboard or from widget according to 'reach' parameter
+      leads_per_status = if w.metadata.leads_per_status && w.metadata.leads_per_status.reach == 'dashboard' then ImpacDashboardsSvc.getCurrentDashboard().leads_per_status else w.metadata.leads_per_status
```
cesar-tonnoir: Should be changed back to status_selection

src/components/widgets/sales-leads-funnel/sales-leads-funnel.tmpl.html
cesar-tonnoir: Hi,
No: the response you will receive from the API won't necessarily match the parameters you send.
For example, the parameter status_selection you send to the API refers to the list of status that you want to take into account. The parameter leads_per_status refers to a list of leads, ordered per status.
So it's not going to be the same.

Please revert back to status_selection

src/components/widgets/sales-opportunities-funnel/sales-opportunities-funnel.directive.coffee
```diff
@@ -34,13 +33,20 @@ module.controller('WidgetSalesOpportunitiesFunnelCtrl', ($scope, $q, ChartFormat
...
+
+      # Parameter which define showing 'Apply to all similar widgets' checkbox
+      $scope.hasReach = true;
+
+      $scope.statusOptions = _.compact _.map(opps_per_sales_stage.values, (status) ->
+        {label: status, selected: true} if angular.isDefined(w.content.opps_per_sales_stage[status]))
 
       angular.forEach w.content.opps_per_sales_stage, (value, status) ->
-        if w.metadata.sales_stage_selection && !(status in w.metadata.sales_stage_selection)
+        if opps_per_sales_stage.values && !(status in opps_per_sales_stage.values)
```
cesar-tonnoir: Please revert back to sales_stage_selection

src/components/widgets/sales-opportunities-funnel/sales-opportunities-funnel.tmpl.html
cesar-tonnoir: Please revert back to sales_stage_selection

src/components/widgets-settings/params-picker/params-picker.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    // Inject services called by the directive...
+    inject(function (_ImpacWidgetsSvc_) {
+      ImpacWidgetsSvc = _ImpacWidgetsSvc_;
+    });
+
+    // ...and stub their methods
+    spyOn(ImpacWidgetsSvc, 'updateWidgetSettings').and.stub();
+  });
+
+  describe('.link', function(){
```
cesar-tonnoir: #setting.toMetadata() is not tested: we should test that the returned hash corresponds to the attributes that have been passed to the directive, correctly formatted.

src/services/widgets/widgets.spec.js
```diff
@@ -56,14 +56,105 @@ describe('<> ImpacWidgetsSvc', function () {
...
+
+    });
+
+    describe('on _self.load success', function () {
+      beforeEach(function() {
+        spyOn(ImpacDashboardsSvc, 'getCurrentDashboard').and.returnValue(dashboard);
+        $rootScope.$apply();
+      });
+
+      it('parameters is defined', function () {
```
cesar-tonnoir: There you are testing that `settings`, that you defined above, has been defined
This test can be removed

src/services/widgets/widgets.spec.js
```diff
@@ -56,14 +56,105 @@ describe('<> ImpacWidgetsSvc', function () {
...
+        expect(dashboard).toBeDefined();
+        expect(settings).toBeDefined();
+      });
+
+      it('expect the param name to be equal "status_selection"', function () {
+        expect(paramName).toBe("status_selection");
```
cesar-tonnoir: Same thing here: you are testing the definition of paramName, that you just defined above

src/services/widgets/widgets.svc.coffee
```diff
@@ -99,29 +99,30 @@ angular
...
-
-          # The widget's metadata are updated only if the correct setting is attached to the widget
-          if _.includes(wgtSettingsKeys, paramName)
-            wgt.metadata[paramName] = _.cloneDeep settings.toMetadata()[paramName]
-            wgt.isLoading = true
-            _self.update(wgt, { metadata: wgt.metadata }).then(
-              (updatedWidget) ->
-                _self.show(updatedWidget).finally( -> updatedWidget.isLoading = false )
-            )
+      _self.load().then ->
```
cesar-tonnoir: just as a curiosity: why did you need _self.load there?

src/services/widgets/widgets.spec.js
```diff
@@ -56,14 +56,105 @@ describe('<> ImpacWidgetsSvc', function () {
...
+      });
+
+      it('expect dashboard have been updated', function () {
+        expect(ImpacDashboardsSvc.update).toHaveBeenCalledWith(dashboard.id, settings.toMetadata());
+      });
+
+      it('expect dashboard setting "reach" will be "dashboard"', function () {
```
cesar-tonnoir: Same thing: it tests the behaviour of DashboardsSvc.update, which has been stubbed above (in your spyOn block)

src/services/widgets/widgets.spec.js
```diff
@@ -56,14 +56,105 @@ describe('<> ImpacWidgetsSvc', function () {
...
+        expect(updatedDashboard[paramName].reach).toBe("dashboard");
+      });
+
+      it('expect update have been called with correct widgets and they metadata', function () {
```
cesar-tonnoir: This one is good, but a test case is missing: it should test that all the similar widgets have been updated as well.
I would split it in two 'it' blocks:

```
  it('updates all the similar widgets')
  it('does not update the other widgets')
```

For that you will need a third widget object, which has a setting that has the same key "status_selection" as widget[0]

=> test 1 will test that widgets[0] and widgets[2] have been updated
=> test 2 will test that widgets[1] has not been updated

src/services/widgets/widgets.spec.js
```diff
@@ -56,14 +56,105 @@ describe('<> ImpacWidgetsSvc', function () {
...
+        expect(svc.update).toHaveBeenCalledWith(dashboard.widgets[0], {metadata: dashboard.widgets[0].metadata});
+        expect(svc.update).not.toHaveBeenCalledWith(dashboard.widgets[1], {metadata: dashboard.widgets[1].metadata});
+      });
+
+      it('expect show have been called', function () {
+        expect(svc.show).toHaveBeenCalled();
```
cesar-tonnoir: good but incomplete: should have been called as many times as widgets have been updated

src/services/widgets/widgets.spec.js
```diff
@@ -56,14 +56,105 @@ describe('<> ImpacWidgetsSvc', function () {
...
+        expect(dashboard).toBeDefined();
+        expect(settings).toBeDefined();
+      });
+
+      it('expect the param name to be equal "status_selection"', function () {
+        expect(paramName).toBe("status_selection");
+      });
+
+      it('expect dashboard have been updated', function () {
+        expect(ImpacDashboardsSvc.update).toHaveBeenCalledWith(dashboard.id, settings.toMetadata());
```
cesar-tonnoir: good one: that's the spirit: you expect the action that has been defined in the function you've implemented / not the variables you use in the spec or the methods that you've stubbed

src/services/widgets/widgets.spec.js
```diff
@@ -56,14 +56,105 @@ describe('<> ImpacWidgetsSvc', function () {
...
+      dashboard.widgets[1].settings = [angular.copy(anotherSettings)];
+
+      paramName = _.keys(settings.toMetadata())[0];
+
+      dashboard[paramName] = {
+        values:["a", "b"],
+        reach: "dashboard"
+      };
+
+      var keys = _.keys(dashboard.widgets[1].settings[0].toMetadata())[0];
```
cesar-tonnoir: this variable is never called?

src/services/widgets/widgets.spec.js
```diff
@@ -56,14 +56,105 @@ describe('<> ImpacWidgetsSvc', function () {
...
+          }
+        }
+      };
+
+      dashboard.widgets[0].settings = [angular.copy(settings)];
+      dashboard.widgets[1].settings = [angular.copy(anotherSettings)];
+
+      paramName = _.keys(settings.toMetadata())[0];
+
+      dashboard[paramName] = {
```
cesar-tonnoir: Are you using this later in the spec? I feel like testing that `DashboardsSvc.update` has been called with `dashboard` should be enough

cesar-tonnoir:
@AterStrix thank you for the modifications on the test files. I am going to pull your branch into my own repository and test the code on my machine today.
I'll keep you posted on the outcome.

PR #156, opened 2016-04-14
Fix/kpis

xaun:
@cesar-tonnoir please review


src/components/kpis-bar/kpis-bar.directive.coffee
```diff
@@ -44,10 +44,7 @@ angular
               $log.error("Impac Kpis bar can't add a kpi", error)
           ).finally(-> $scope.isAddingKPI = false)
 
-        $scope.removeKpi = (kpiId) ->
-          $scope.kpis = _.remove($scope.kpis, (kpi) ->
-            kpi.id != kpiId
```
xaun: @alexnoox this line you added caused all kpi's other than the one you wanted to delete to be erased :P

PR #160, opened 2016-04-18
Fix #IMPAC-172

x4d3:
This PR is linked with https://github.com/maestrano/impac/pull/255

Fix from https://maestrano.atlassian.net/browse/IMPAC-172
2. Add time period selection capabilities (will require to modify connec! report to enable time filtering on opportunities)
3. Add setting "params picker" to help determine which statuses have to be considered as "closed won" and "closed lost" (see widget "leads funnel"). Any status including the mention "won" (eg: "Closed won") will be picked by default, same for statuses including "lost".
5. Enable selection of multiple lines in the breakdown (see widget "profit and loss")


src/components/widgets-settings/params-picker/params-picker.directive.coffee
```diff
@@ -39,9 +38,11 @@ module.directive('settingParamsPicker', ($templateCache) ->
       deferred: '='
       param: '@',
       options: '=',
+      description: '@',
```
cesar-tonnoir: should be '@?' to let the developers use the directive without having to specify a custom description

x4d3:
Closing because it is on the wrong branch.

PR #174, opened 2016-05-06
Team performance widget update

cesar-tonnoir:
add time range to sales performance widget
Add setting to determine which statuses have to be considered as "closed won" and "closed lost"
made description optionnal
Enable selection of multiple lines in the breakdown
add forecast values
make a message appear if you don't select the stages_status


cesar-tonnoir:
@x4d3 removing the dist files should remove the conflicts
have a look at the changelog of v1.3.1 to fix the issues with the params-picker

PR #177, opened 2016-05-16
Pusher Websockets - inapp notifications

xaun:
Not ready for merge, need to test a couple things tomorrow.

@cesar-tonnoir :beer:

Ps, you committed you `workspace/index.js` local configurations -_- lol


x4d3:
What would be awesome, is to add a index.local.js, in .gitignore, and first try with the .local and if not found, use the index.js

xaun:
@x4d3 I bet ya Cesar would still commit that :P I cant find a good way to load a `.js` file in es5 without http. I don't really think it's worth doing any of this rubbish: http://stackoverflow.com/questions/950087/include-a-javascript-file-in-another-javascript-file

src/services/alerts/alerts.svc.coffee
```diff
@@ -0,0 +1,75 @@
...
+  .service('ImpacAlerts', ($log, $q, $http, $httpParamSerializerJQLike, ImpacRoutes, ImpacMainSvc, ImpacEvents, IMPAC_EVENTS)->
+
+    _self = @
+
+    #====================================
+    # Getters
+    #====================================
+    @config = {}
+
+    @config.ssoSessionId = ""
```
cesar-tonnoir: as discussed / sso_session shouldn't be used there:
- api/v2 will authenticate using basic auth
- js_api/v1 will use Devise

src/services/events/events.svc.coffee
```diff
@@ -0,0 +1,26 @@
+
```
cesar-tonnoir: cool stuff

src/services/kpis/kpis.svc.coffee
```diff
@@ -9,19 +9,33 @@ angular
     #====================================
 
     @config = {}
+    @config.userData = {}
 
-    @config.ssoSessionId = ""
```
cesar-tonnoir: :+1:

src/services/kpis/kpis.svc.coffee
```diff
@@ -9,19 +9,33 @@ angular
...
+    @config.currentDhb = {}
+    @getCurrentDashboard = ->
+      return _self.config.currentDhb
+
 
+    #====================================
+    # Register Listeners
+    #====================================
+    ImpacEvents.registerCb(IMPAC_EVENTS.kpiTargetAlert, (notification) ->
+      _self.refreshAll(true)
```
cesar-tonnoir: you may want to refresh another dashboard (which is not currentDashboard)?
`dashboardId` could be a param included in `notification`, then svc.refreshAll(dashboardId)

src/services/kpis/kpis.svc.coffee
```diff
@@ -42,10 +56,12 @@ angular
...
       deferred = $q.defer()
 
-      if _.isEmpty(_self.config.ssoSessionId) || force
+      if _.isEmpty(_self.getSsoSessionId()) || force
         ImpacMainSvc.loadUserData(force).then(
           (mainConfig) ->
-            _self.config.ssoSessionId = mainConfig.sso_session
+            _self.config.userData.id = mainConfig.id
+            _self.config.userData.orgUids = _.map(mainConfig.organizations, (o)-> o.uid)
+            _self.config.userData.ssoSessionId = mainConfig.sso_session if mainConfig.sso_session
```
cesar-tonnoir: Do we really need this? If we have a `getSsoSessionId()` method that is used everywhere in this service, it could just return `MainSvc.config.sso_session` so that the sso session is stored at only one place...

src/services/kpis/kpis.svc.coffee
```diff
@@ -42,10 +56,12 @@ angular
     @load = (force=false) ->
       deferred = $q.defer()
 
-      if _.isEmpty(_self.config.ssoSessionId) || force
+      if _.isEmpty(_self.getSsoSessionId()) || force
         ImpacMainSvc.loadUserData(force).then(
           (mainConfig) ->
-            _self.config.ssoSessionId = mainConfig.sso_session
+            _self.config.userData.id = mainConfig.id
+            _self.config.userData.orgUids = _.map(mainConfig.organizations, (o)-> o.uid)
```
cesar-tonnoir: No a big fan.. where is that used?

src/services/kpis/kpis.svc.coffee
```diff
@@ -62,15 +78,15 @@ angular
       # Re-assign a new initialization promise, if the previous has been resolved or rejected.
       _self.initialized = $q.defer() if _self.initialized.promise.$$state > 0
       _self.load().then( ->
-        _self.config.currentDashboardId = dashboard.id
+        _self.config.currentDhb = dashboard
```
cesar-tonnoir: Why?

src/services/kpis/kpis.svc.coffee
xaun: @cesar-tonnoir `@getAlerts`, I only need the user id.. but we may also need the user orgUids later for org wide alerts

src/services/kpis/kpis.svc.coffee
```diff
@@ -122,8 +138,18 @@ angular
       promises = []
 
       createUrl = ImpacRoutes.kpis.alerts.create(kpi.id)
+
+
       for alert in alertsToCreate
-        promises.push $http.post(createUrl, {alert: _.pick(alert, ['service'])})
+        # Configure alert for inapp websockets with Pusher.
```
cesar-tonnoir: I really think this should be part of the Alert model in mno-hub, upon creation of an 'inapp' alert...

src/services/kpis/kpis.svc.coffee
```diff
@@ -9,19 +9,33 @@ angular
...
     @config.kpisTemplates = []
     @getKpisTemplates = ->
       return _self.config.kpisTemplates
 
-    @config.currentDashboardId = ""
-    @getCurrentDashboardId = ->
-      return _self.config.currentDashboardId
+    @config.currentDhb = {}
+    @getCurrentDashboard = ->
+      return _self.config.currentDhb
```
cesar-tonnoir: could return `ImpacDashboardsSvc.getCurrentDashboard()` => would avoid to store `currentDhb` in the KpisSvc

src/services/kpis/kpis.svc.coffee
```diff
@@ -141,6 +167,16 @@ angular
...
             else
               kpi.alerts.push resp.data
+          ImpacEvents.notifyCallbacks(IMPAC_EVENTS.addOrRemoveAlerts)
+      )
+
+    @refreshAll = (refreshCache=false) ->
+      _self.load().then(->
+        for k in _self.getCurrentDashboard().kpis
+          _self.show(k, refreshCache).then(
+            (renderedKpi)-> $log.debug('ImpacKpisSvc: Successfully refreshed Kpis.')
```
cesar-tonnoir: plz remove debug log

src/services/kpis/kpis.svc.coffee
xaun: `@refreshAll`

src/services/pusher/pusher.svc.coffee
```diff
@@ -0,0 +1,66 @@
...
+# ---------------------------------------------------------------
+
+angular
+  .module('impac.services.pusher', [])
+  .service('Pusher', ($window) ->
+
+    _self = @
+
+    @config =
+      key: 'e98dfd8e4a359a7faf48'
```
cesar-tonnoir: store in config

src/services/kpis/kpis.svc.coffee
cesar-tonnoir: again, as we've been talking about getters, I would prefer getters that call configs of underlying services instead of storing the objects twice

src/services/kpis/kpis.svc.coffee
cesar-tonnoir: couldn't we `@refreshAll(dashboardId)` instead?

PR #180, opened 2016-05-26
Fix issue when user has less than 3 organization

x4d3:


cesar-tonnoir:
commit cherry-picked to release/1.3.5

PR #188, opened 2016-06-12
Feature/ch widgets

p-mob2015:
Pull request for "Growth Widget for Product"
- Classic selector
- Name length coordinator


cesar-tonnoir:
Hi @p-mob2015,
Thanks for your pull request - I'll add a few inline comments.

src/components/widgets-settings/param-selector-classic/param-selector-classic.directive.coffee
```diff
@@ -0,0 +1,60 @@
+module = angular.module('impac.components.widgets-settings.param-selector-classic', [])
```
cesar-tonnoir: This directive is quite similar to the params-selector we already have: don't you think we could add a new parameter to the existing directive instead? Like:
```
scope: {
  ...
  classic: '=?'
}
```

A test on the value of $scope.classic (in the template) could determine if you should display the usual container or a simple <select>.

src/components/widgets/sales-growth/sales-growth.directive.coffee
```diff
@@ -26,10 +28,12 @@ module.controller('WidgetSalesGrowthCtrl', ($scope, $q, ChartFormatterSvc, $filt
   $scope.isDataQuantity = true
   w.initContext = ->
     if $scope.isDataFound = angular.isDefined(w.content) && !_.isEmpty(w.content.summary) && !_.isEmpty(w.content.dates)
-
+      
       $scope.productOptions = _.flatten(_.map(w.content.summary, (product) ->
-        return {label: product.code, value: product.id}
+        label = if w.content.organizations.length = 1 then product.name else product.company + ' - ' + product.name
```
cesar-tonnoir: The test on threshold should also be applied here. Otherwise, you may end-up with products with a huge name directly in the <select>. Unless maybe the name is automatically truncated when you use select?

cesar-tonnoir:
merged with 1.3.8 (see https://github.com/maestrano/impac-angular/pull/197)

PR #190, opened 2016-06-16
Changes for updated kpi api response to support extra_watchables

xaun:


xaun:
https://github.com/maestrano/impac/pull/308

src/components/kpi/kpi.directive.coffee
```diff
@@ -36,11 +36,18 @@ angular
             if kpiTemplate? && kpiTemplate.extra_params?
               $scope.kpi.possibleExtraParams = kpiTemplate.extra_params
 
-            $scope.kpi.targets ||= []
-            if !_.isEmpty($scope.kpi.targets[0])
+            $scope.kpi.data ||= {}
+            $scope.kpi.results ||= {}
+            $scope.kpi.targets ||= {}
+
+            $scope.watchableData = $scope.kpi.data[$scope.kpi.element_watched]
```
cesar-tonnoir: have you tested that when you edit the settings? because on mine it's broken (wouldn't it be better to keep reading directly the kpi object?).
I think that's because when you update the settings, it calls UPDATE, which calls SHOW, which does not update $scope.watchableData

PR #201, opened 2016-07-03
[IMPAC-175] [IMPAC-286] Kpis selector

cesar-tonnoir:
@xaun can you please review?

Status update - still a few things to fix:
- remove account name from kpi attached to widget
- discovery endpoint wasn't working on mine?..
- saving the kpis when on edit mode is not obvious
- alerting modal should take all the watchables into account
- API to return dynamic edit captions
- API to return appropriate emphasis for all the existing kpis
- kpis bar's `kpisEditSettings` is buggy: settings are not removed when the kpi is removed from `$scope.kpis` -> couldn't the edit settings be saved in the kpi object (in $scope.kpis)
- bug spotted (not appearing 100% of the time): when I try to attach a kpi "too fast", I receive a 400 because the element_watched cannot be blank. Probably a problem with the ng-init: we should attach the element_watched to the object directly in the service when we retrieve the list of the available kpis (until we implement a better logic)
- metadata hist_parameters is incorrect (we should add at least a dates-picker to let the user select the time range)

### DIFF
--- a/src/components/kpis-bar/kpis-bar.directive.coffee
+++ b/src/components/kpis-bar/kpis-bar.directive.coffee
@@ -11,7 +11,11 @@ angular
       controller: ($scope, $timeout, $log) ->
         # Load
         # -------------------------
-        $scope.hideAvailableKpis = true
+        $scope.availableKpis = {
+          hide: true,
+          toggle: ->
+            $scope.availableKpis.hide = !$scope.availableKpis.hide
+        }
         $scope.showKpisExpanded = false
         # All kpis edit panels are shown
         $scope.showEditMode = false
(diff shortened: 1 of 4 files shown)


src/components/kpis-bar/kpis-bar.directive.coffee
```diff
@@ -53,11 +57,16 @@ angular
...
         }
 
+        initAvailableKpis = ->
+          $scope.availableKpis.list = _.select ImpacKpisSvc.getKpisTemplates(), (k) ->
+            _.isEmpty(k.attachables) &&
+            _.isEmpty(_.select($scope.kpis, (existingKpi) ->
+              existingKpi.endpoint == k.endpoint && existingKpi.element_watched == k.watchables[0]
+            ))
+          $scope.availableKpis.hide = true if _.isEmpty($scope.availableKpis.list)
+
```
xaun: note: maybe look at using slice with a removeKpi method

PR #204, opened 2016-07-08
[IMPAC-303] - KPI FE - Drag & drop now working

clemthenem:
### Fixed bug
Fixed the issue with the drag and drop on KPIs not working.

### DIFF
--- a/src/components/kpis-bar/kpis-bar.tmpl.html
+++ b/src/components/kpis-bar/kpis-bar.tmpl.html
@@ -45,7 +45,7 @@
     <div class="col-xs-12 col-sm-12">
       <div class="row">
         <div ui-sortable="sortableOptions" ng-model="kpis">
-          <div ng-repeat="kpi in kpis" impac-kpi class="impac-kpi col-xs-6 col-sm-4 col-md-3" kpi="kpi" on-delete="removeKpi(kpi.id)" edit-mode="showEditMode" available-kpis="availableKpis.list">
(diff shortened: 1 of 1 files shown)


src/components/kpis-bar/kpis-bar.tmpl.html
```diff
@@ -45,8 +45,7 @@
     <div class="col-xs-12 col-sm-12">
       <div class="row">
         <div ui-sortable="sortableOptions" ng-model="kpis">
-          <div ng-repeat="kpi in kpis track by kpi.id">
-            <div impac-kpi class="impac-kpi col-xs-6 col-sm-4 col-md-3" kpi="kpi" on-delete="removeKpi(kpi.id)" edit-mode="showEditMode" available-kpis="availableKpis.list" />
+          <div ng-repeat="kpi in kpis" impac-kpi class="impac-kpi col-xs-6 col-sm-4 col-md-3" kpi="kpi" on-delete="removeKpi(kpi.id)" edit-mode="showEditMode" available-kpis="availableKpis.list">
```
xaun: @clemthenem why have you removed the `track by kpi.id`? If you use `track by` does it really stop this from working?

clemthenem: No it doesn't you are right! I'll fix this

xaun:
@cesar-tonnoir this should be good to merge.

The change I made to move it into two divs was when I was experimenting with the clearfix rows when adding kpis. 

You can see the commit here: https://github.com/maestrano/impac-angular/commit/3104c13b336fd7eec88360b8ba254d6cbe333180, the method I added to the directive does not exist anymore, but I forgot to revert the template changes - which is probably why the drag & drop got broken in the first place =\, so clems fix is exact. 

:beers:

PR #205, opened 2016-07-11
[IMPAC-277] - Sortable the drill down widgets (1)

an-pavlovych:


cesar-tonnoir:
@an-pavlovych it's working well and matches the requirements: :+1: 
@xaun: review is ok for me, but I think we should add a jira somewhere to wrap the sorting logic in a "common" directive.
I've done a few minor changes there: https://github.com/maestrano/impac-angular/pull/231/commits/38b69c794c4cdf990c6bd0021affe64a71026cde

xaun:
@cesar-tonnoir yeah looks good to me. Definitely looks like a good idea to move the sorting logic into a directive.

PR #209, opened 2016-07-15
[IMPAC-276] Changed view Account Balance to collapsable list (3)

an-pavlovych:


cesar-tonnoir:
Hi @an-pavlovych,
I think there has been a misunderstanding there: The idea was just to group the accounts in the drop-down by classification, not to transform the whole widget to a "drill-down" widget... :/

Your change in the API (https://github.com/maestrano/impac/pull/321/files) is ok and is going to be merged, but this PR has to be reworked: why wouldn't you simply keep the drop-down list and add the classification name in front of the account name?

Like for example: "Accounts Payable" would be become "[Asset] Accounts Payable".
In case of a multi-company dashboard, "Maestrano - Accounts Payable" would be "Maestrano - [Asset] Accounts Payable".
Then, all the "Asset" accounts should be grouped in the list, same thing for the other classifications.

Does that make more sense to you?

an-pavlovych:
I just added group by option to selecting the acc

cesar-tonnoir:
@an-pavlovych, yes that's exactly what we needed, I didn't even know it could be that simple :)
change included in https://github.com/maestrano/impac-angular/pull/236

PR #216, opened 2016-07-19
[IMPAC-117] Select time period on aged sales and number of leads (7)

an-pavlovych:


cesar-tonnoir:
The `dates-picker` setting has been used, but as specified on the jira card, we need the `time-period-selection` setting there: indeed, the users are more likely to use the presets ("last 6 months", "previous financial year", etc.) on these widgets rather than simply selecting two dates

PR #221, opened 2016-07-22
[IMPAC-279] Add hist mode to expense widgets (8)

an-pavlovych:


src/components/widgets/accounts-expense-weight/accounts-expense-weight.less
```diff
@@ -1,4 +1,6 @@
 .analytics .widget-item .content.accounts-expense-weight {
+  height: 285px;
```
cesar-tonnoir: warning: forcing the widget's height will cause th grid system to break

cesar-tonnoir:
slightly reworked and included in https://github.com/maestrano/impac-angular/pull/236

PR #222, opened 2016-07-26
Fixed a little bug with the info-pannel

clemthenem:
Fixed a little bug with the info-pannel.
The issue was that when a user clicked the setting button and then the info button, the info was partially hidden by the settings.
E.g: Account Balance widget.


cesar-tonnoir:
good pickup, thanks!

PR #223, opened 2016-07-26
[IMPAC-263] Added hist mode to custom calculation (9)

an-pavlovych:
In this task I've added `hist mode choser` as directive to `accounts-custom-calculation.tmpl.html`. Also added `setting-time-period` directive to template with widget settings. Though I faced the problem: these two directives some how create different `hist_parameters` objects in `widget.settings`. Hist mode switcher goes broken because of it. I.e. `hist mode` works till we open settings. Then `setting-time-period` addes new object `hist_parameters` to `widget.settings`, and do not use created. For now I cannot find the reason of this bug.


src/components/widgets-settings/accounts-list/accounts-list.directive.coffee
```diff
@@ -33,9 +33,8 @@ module.controller('SettingAccountsListCtrl', ($scope, $timeout, ImpacWidgetsSvc)
 
     if w.content? && !_.isEmpty(w.content.complete_list)
       w.remainingAccounts = angular.copy(w.content.complete_list)
-      $timeout () ->
```
cesar-tonnoir: @an-pavlovych the timeout was there to ensure all the context is passed to the directive before executing the next statements.
Any particular reason for removing it?

src/components/widgets-settings/formula/formula.directive.coffee
```diff
@@ -15,9 +15,8 @@ module.controller('SettingFormulaCtrl', ($scope, $filter, $timeout) ->
   setting.initialize = ->
     if w.metadata? && w.metadata.formula?
       w.formula = w.metadata.formula
-      $timeout () ->
```
cesar-tonnoir: @an-pavlovych same here ?

an-pavlovych:
I've removed async there in order to have data calculated by formula in `WidgetAccountsCustomCalculationCtrl` directove in `w.format` method when creating the chart.

src/components/widgets-settings/formula/formula.directive.coffee
```diff
@@ -65,13 +68,25 @@ module.controller('SettingFormulaCtrl', ($scope, $filter, $timeout) ->
...
+    return evaluatedFormula
+
   formatFormula = ->
     if !w.formula.match(/\//g) && w.selectedAccounts?
       if firstAcc = w.selectedAccounts[0]
         if currency = firstAcc.currency
           w.evaluatedFormula = $filter('mnoCurrency')(w.evaluatedFormula, currency)
 
-  w.settings.push(setting)
+  # w.settings.push(setting)
```
cesar-tonnoir: the widget will never load its content if you remove this line..

src/components/widgets-settings/formula/formula.directive.coffee
```diff
@@ -48,15 +48,18 @@ module.controller('SettingFormulaCtrl', ($scope, $filter, $timeout) ->
...
-    if (!str.match(authorized_regex))
-      w.isFormulaCorrect = false
-      w.evaluatedFormula = "invalid expression"
+    w.evaluatedFormula = applyFormula(str)
 
-    try
-      w.evaluatedFormula = eval(str).toFixed(2)
-    catch e
-      w.evaluatedFormula = "invalid expression"
+    if w.selectedAccounts? && firstAcc = w.selectedAccounts[0]
```
cesar-tonnoir: @an-pavlovych code duplication here. The string replacement could have been put in a function and re-used

cesar-tonnoir:
changes + corrections included in https://github.com/maestrano/impac-angular/pull/241

PR #224, opened 2016-07-27
[IMPAC-57] Aged payables and receivables to display cumulative chart (10)

an-pavlovych:


cesar-tonnoir:
@an-pavlovych: Code is ok for me, but needs to be wrapped into a `widget-setting` directive, as per the requirements in the jira card: we will reuse this directive in other widgets with line charts.

PR #232, opened 2016-07-29
[IMPAC-275] Added legend to pie charts

an-pavlovych:


src/components/widgets/accounts-assets-liability-summary/accounts-assets-liability-summary.less
```diff
@@ -1,15 +1,15 @@
 .analytics .widget-item .content.accounts-assets-liability-summary {
+  height: 300px;
```
cesar-tonnoir: @an-pavlovych again, please be careful with that: the widgets' height must not be changed or the all grid system will be broken.
I'll correct it for this PR.

src/components/widgets/sales-cycle/sales-cycle.directive.coffee
```diff
@@ -45,17 +47,23 @@ module.controller('WidgetSalesCycleCtrl', ($scope, $q, ChartFormatterSvc, $filte
     if error.code == 404
       $scope.isDataFound = false
 
+  $scope.getAccountColor = (elem) ->
```
cesar-tonnoir: naming: `getAccountColor` doesn't make sense in a sales widget

src/components/widgets/sales-cycle/sales-cycle.directive.coffee
```diff
@@ -45,17 +47,23 @@ module.controller('WidgetSalesCycleCtrl', ($scope, $q, ChartFormatterSvc, $filte
...
+    if name?
+      return name.replace(/_/g, " ")
 
   # Chart formating function
   # --------------------------------------
   $scope.drawTrigger = $q.defer()
   w.format = ->
     if $scope.isDataFound
-      pieData = _.compact _.map $scope.statusOptions, (statusOption) ->
+      $scope.pieData = _.compact _.map $scope.statusOptions, (statusOption) ->
```
cesar-tonnoir: why adding this `pieData` variable to the `$scope` there? do you need it in the template?

src/components/widgets/sales-summary/sales-summary.directive.coffee
```diff
@@ -66,11 +66,13 @@ module.controller('WidgetSalesSummaryCtrl', ($scope, $q, ChartFormatterSvc) ->
         tooltipFontSize: 12,
       }
       angular.merge(pieOptions, {currency: 'hide'}) if $scope.filter.value.indexOf('quantity') > -1
-      chartData = ChartFormatterSvc.pieChart(pieData, pieOptions)
+      chartData = ChartFormatterSvc.pieChart($scope.pieData, pieOptions)
 
       # calls chart.draw()
       $scope.drawTrigger.notify(chartData)
 
+  $scope.getAccountColor = (elem) ->
```
cesar-tonnoir: same as above: be careful of the naming of your variables/methods

cesar-tonnoir:
Hi @an-pavlovych,

I've been testing this code: we're almost there! The legends are showing the correct data, but I cannot merge it as it is because you've modified the height of the different widgets, which is breaking the bootstrap grid we are using:

voila_capture 2016-08-11_11-29-58_am

So here is what I propose:
- For the "small" widgets (assets/liabilities summary, assets summary and expenses/revenue): please keep the height as it was before. I know the legend don't look awesome (like 1 line will be visible and you'll have to scroll to see the others), but it will change when we update our grid system to Gridster (soon)
- For the "wide" widgets (top customers summary, sales summary/by location/by industry/by customer, sales cycle): use the less mixin `.tall-widget()` (https://github.com/maestrano/impac-angular/blob/master/src/stylesheets/mixins.less#L48) so that these widgets will have the same height as Profit and loss, balance sheet, etc. Eventually, you can also create a new "master style" (https://github.com/maestrano/impac-angular/blob/master/src/stylesheets/widget-master-styles.less#L27)

Let me know if you need more details.

PR #237, opened 2016-08-04
[IMPAC-65] Added sales/opportunity_timeline widget

an-pavlovych:
{
        path: 'sales/opportunity_timeline',
        name: 'Opportunity timeline',
        metadata: { template: 'sales/opportunity_timeline' },
        desc: 'Opportunity timeline',
        icon: 'line-chart',
        width: 6
 }


xaun:
The code you have written looks good, just I think there is a bit of a misunderstanding with the concept of the widget.

Below is a list of changes to be made to bring this widget inline with the requirements. It will require changes to both the engine and the front-end.

- The left panel opportunities list:
  - List opportunities in a **non-expandable** list **by Opportunity name** (no need to display the sales stages)
  - Selected opportunities should be visibly selected (See the "Sales Comparison" widget for example), currently you cannot tell which item is selected from the left panel. See image below:
screen shot 2016-08-18 at 12 36 13 pm
  - A maximum of 1 opportunity may be selected at any given time
  - There should also be an "average opportunities" entry added to the opportunities array (which can be selected and displayed on the pie chart). This should be created in the engine by determining the averages of each opportunity.
  - The "Average Opportunities" opportunity will need to contain a sales stages array, with averages calculated from all the opportunities sales stage's across the opportunities.

- The right panel graph section:
  - The graph should display only 1 opportunity at any given time.
  - The graph sections (slices) should represent each sales stage of the selected opportunity
  - The list below the graph should list the selected Opportunity's sales stages like so: sales stage name, sales stage date range, sales stage total duration. So in other words, each slice of the pie graph (being the sales stages for the selected Opportunity).
  - The heading above the graph should be the opportunity name
  - The sub-heading below the graph can be removed

- In the settings panel:
  - Remove the time period selection directive.
  - Take a look at the "Opportunities Funnel" widget settings, the "Sales Stage Selection" section I believe uses the params-picker directive. That should be used in the widgets settings to select which sales stages the User wants to compare. See image below:
screen shot 2016-08-18 at 12 36 41 pm

Thanks

PR #244, opened 2016-08-18
Release/1.3.13

xaun:
@cesar-tonnoir @alexnoox Please review.


src/components/widgets-common/data-not-found/data-not-found.directive.coffee
```diff
@@ -22,9 +23,11 @@ module.directive('commonDataNotFound', ($templateCache, $log, $http, ImpacAssets
 
         image.onerror = ->
           $log.warn("Missing data-not-found image for #{scope.widgetEngine}")
+          debugger
```
cesar-tonnoir: warning: `debugger`s to be removed

xaun: You're commenting on the diff from clairs commit I think?

src/components/widgets-common/data-not-found/data-not-found.directive.coffee
```diff
@@ -5,26 +5,38 @@ module.directive('commonDataNotFound', ($templateCache, $log, $http, ImpacAssets
...
+      else
+        usingDefaults = true
+        imagePath = 'dist/images/widget-bg-width-' + scope.widgetWidth + '.png'
 
-        image.onerror = ->
-          $log.warn("Missing data-not-found image for #{scope.widgetEngine}")
-          image.remove() if image?
+      image.onerror = ->
+        missingImageLocationMsg = if usingDefaults then 'library defaults' else scope.widgetEngine
+        $log.warn("Missing data-not-found image for #{missingImageLocationMsg}")
```
cesar-tonnoir: @xaun I would indicate the whole image path, otherwise it can become confusing if a default image has been provided for another widget size.
Couldn't we provide defaults image directly with the library? So that worst case scenario, the "maestrano" defaults images are displayed?

xaun: @cesar-tonnoir that is exactly what is happen, impac-angular is providing default images within the library, but not a full suite of images, just one for each width.

workspace/index.js
```diff
@@ -119,11 +119,6 @@ module.run(function ($log, $q, $http, ImpacLinking, ImpacAssets, ImpacRoutes, Im
     }
   });
 
-  // Link Impac! Assets
```
cesar-tonnoir: why is that removed?

src/components/widgets-common/data-not-found/data-not-found.directive.coffee
cesar-tonnoir: That was my impression of the gulp task, but then I'm confused: in which case could we have the "missing data not found" warning?

workspace/index.js
xaun: Because the library is providing it by default - the whole point of these changes

cesar-tonnoir: ok, I missed this one: 
workspace/assets/impac-logo.png → src/images/impac-title-logo.png
:+1:

src/components/widgets-common/data-not-found/data-not-found.directive.coffee
xaun: case 1. you have provided a custom data_not_found directory in your parent application, and the image for that widget is not found.
case 2. you have not provided a custom data_not_found directory in your parent application, and for some reason, the assets that are meant to be provided are missing from impac-angular/dist

cesar-tonnoir: ok for the case 2.,
but the case 1. should fallback to the default image. Otherwise, when you start uploading custom screenshots for a few widgets, you're forced to add one for each of the widgets.

xaun: @cesar-tonnoir I agree, but I was having trouble with writing the code for that one and I thought that at least this would be a good start and that that logic could be further improved in another minor release as I was trying to get it out for noox. If you think the time should be spent for this release let me know.

alexnoox:
LGTM

src/components/widgets-common/data-not-found/data-not-found.directive.coffee
```diff
@@ -26,17 +35,16 @@ module.directive('commonDataNotFound', ($templateCache, $log, $http, ImpacAssets
 
         imagePath = dir + scope.widgetEngine + '.png'
 
+        # Check if custom image has been provided, if not, display default.
+        $http.get(imagePath).then(
```
cesar-tonnoir: :+1:

cesar-tonnoir:
LGTM2

PR #245, opened 2016-08-18
[IMPAC-325] Simplifies sync-apps directive

cesar-tonnoir:
+ Display retrieved statuses rather than interpretations
+ Constant polling and state update
+ Refresh dashboard on syncing state change
+ "Show status" instead of "is synced" / "not synced"
+ Statuses available while syncing


cesar-tonnoir:
@xaun can you review please? thx!

src/components/dashboard-settings/sync-apps/sync-apps.directive.js.coffee
```diff
@@ -89,108 +40,73 @@ module.directive('dashboardSettingSyncApps', ($templateCache, $log, $http, $filt
...
-              list.find('.message').addClass('ng-hide')
-              listItem.find('.message').removeClass('ng-hide')
-              # Calculates the top offset of the listItem relative to the list.
-              list.animate({ scrollTop: listItem.offset().top - list.offset().top }, 500)
-              return
+          controller: ($scope, connectors) ->
+            $scope.connectors = connectors
+            $scope.expandListItemOnClick = (connector) ->
+              return unless connector.message
+              connector.showMessage = !!!connector.showMessage
```
xaun: Casting `connector.showMessage` into a boolean then negating the boolean could be simplified with just `!connector.showMessage`. The result will be the same it seems.

xaun:
Look good to me, clean code good comments 👍

src/components/dashboard-settings/sync-apps/sync-apps.directive.js.coffee
xaun: Thinking about this further, I don't think I mind it as it would be consistent with needing `connector.showMessage` to be `true` in the case of assigning to a variable

cesar-tonnoir: Weird what I've done there... `!!!variable` is equivalent to `!variable` isn't it? `!null == true`...

xaun: variable = 'hello'
!variable == false #=> true
variable == true #=> false

cesar-tonnoir:
@xaun can you merge please?

PR #249, opened 2016-08-29
Kpis fixes

xaun:


src/components/kpi/kpi.directive.coffee
```diff
@@ -98,11 +111,17 @@ angular
         $scope.hasValidTargets = ->
           ImpacKpisSvc.validateKpiTargets($scope.targets)
 
-        $scope.updateSettings = ->
+        # Prevent displaying text characters before expresses are loaded.
+        $scope.buildEphasis = (kpi)->
```
cesar-tonnoir: I get why you're doing that here, but not a big fan: this belongs to the view...
Maybe simply wrap it into a `div` or `span` with a `ng-if` condition?

src/components/kpi/kpi.directive.coffee
```diff
@@ -98,11 +111,17 @@ angular
...
+          return unless kpi && kpi.layout && kpi.layout.text && kpi.data && kpi.watchables.length
+          $sce.trustAsHtml("<strong>#{kpi.layout.text.emphasis}</strong> - <span class=\"real-value\">(#{kpi.data[kpi.watchables[0]].value} #{kpi.data[kpi.watchables[0]].unit})</span>")
+
+        $scope.updateSettings = (force)->
```
cesar-tonnoir: when do you need to force it?

src/components/kpi/kpi.tmpl.html
```diff
@@ -45,7 +44,7 @@
         <div class="im-form-group">
           <div ng-repeat="(param, paramValues) in kpi.possibleExtraParams track by $index">
             <!-- extra param selection -->
-            <select name="extraParam" ng-model="kpi.extra_params[param]" ng-options="value.id as value.label for value in paramValues" ng-init="kpi.extra_params[param] = paramValues[0].id">
```
cesar-tonnoir: :+1:

cesar-tonnoir:
ok

PR #251, opened 2016-09-06
[IMPAC-65] Opportunities timeline widget

an-pavlovych:


an-pavlovych:
path: 'sales/opportunities_timeline',
        name: 'Opportunities timeline',
        metadata: { template: 'sales/opportunities_timeline' },
        desc: 'Opportunites timeline',
        icon: 'line-chart',
        width: 6

src/components/widgets/sales-opportunities-timeline/sales-opportunities-timeline.directive.coffee
```diff
@@ -0,0 +1,151 @@
...
+
+  $scope.sort = (col) ->
+    if $scope.sortedColumn == col
+      $scope.ascending = !$scope.ascending
+    else
+      $scope.ascending = true
+      $scope.sortedColumn = col
+    sortData()
+
+  $scope.filterBySalesStage = (element) ->
```
xaun: @an-pavlovych this type of filtering should be done in the API. 

When a user updates the widgets-settings, it will post to maestrano-api saving the widget.metadata, then a get_widget request will be made to Impac API. You should detect these metadata settings in Impac, and return the filtered results rather than using JS and $filter.

src/components/widgets/sales-opportunities-timeline/sales-opportunities-timeline.directive.coffee
```diff
@@ -0,0 +1,151 @@
...
+        ImpacWidgetsSvc.updateWidgetSettings(w,false)
+
+  $scope.isSelected = (element) ->
+    element? && $scope.selectedElement? && (
+      matcher = (if element.id? then 'id' else 'name')
+      $scope.selectedElement[matcher] == element[matcher]
+    )
+
+  # <---
+
```
xaun: Some comments for the column sorting code here would be nice

src/components/widgets/sales-opportunities-timeline/sales-opportunities-timeline.tmpl.html
```diff
@@ -0,0 +1,95 @@
...
+
+        <div ng-hide="selectedElement" class="no-element">
+          Select one or several employee(s) or category(ies) to display the corresponding summary.
+        </div>
+
+      </div>
+
+    </div>
+
+    <div ng-show="(isDataFound==false)" common-data-not-found widget-engine="widget.category" />
```
xaun: Please update this to `<div ng-show="(isDataFound==false)" common-data-not-found widget-engine="widget.category" width="widget.width"/>`, we have added a new feature recently that needs the widgets width passed into the `common-data-not-found` directive.

xaun:
@an-pavlovych the sales stage date range return an empty string when there is no range. This should be handled more intelligently to prevent the poor display in the DOM like the image below.

screen shot 2016-09-08 at 11 43 33 am

Also, could you please change the formatting to something nicer, like  "11/07/2016 to 03/08/2016" and when there is only one date, just "03/08/2016".

src/components/widgets/sales-opportunities-timeline/sales-opportunities-timeline.tmpl.html
```diff
@@ -0,0 +1,95 @@
+<div widget-sales-opportunities-timeline>
+
+  <div class="edit" ng-show="widget.isEditMode">
+    <h4>Widget settings</h4>
+    <div setting-organizations parent-widget="widget" class="part" deferred="::orgDeferred" />
+    <div setting-params-picker options="statusOptions" has-reach="hasReach" param="sales_stage_selection" parent-widget="widget" class="part" deferred="::paramsPickerDeferred1" />
+    <div setting-params-picker options="assigneesOptions" has-reach="hasReach" param="assignees_selection" parent-widget="widget" class="part" deferred="::paramsPickerDeferred2" />
```
xaun: @an-pavlovych please remove this widget setting (assignees_selection), it is not needed.

PR #252, opened 2016-09-16
[IMPAC-333] Added the toggle a hide/show effect on the whole KPIs bar

an-pavlovych:


src/components/kpis-bar/kpis-bar.less
```diff
@@ -83,52 +64,99 @@
...
+        }
+      }
     }
-  }
 
-  .add-bar {
-    border-bottom: dashed 1px;
-  }
+    &.hide-animate {
+      min-height: 0;
```
xaun: This seems to have no effect, I think you're looking for `max-height`, but even that will not reduce the height of the kpi-bar and all it's children to 0.

src/components/kpis-bar/kpis-bar.less
```diff
@@ -83,52 +64,99 @@
...
+      .sdp-to-label {
+        text-transform: uppercase;
+        font-weight: 400;
+      }
+    }
+  }
+
+  .show-animate {
+    transition: opacity 1s ease-in;
+    opacity: 1;
```
xaun: Opacity (and especially opacity alone) is not the best choice here... Im not sure if you have noticed, but once you hide the kpis-bar you can still click on the buttons even though they are not visible, they are still in the DOM. The whole kpis-bar and kpis, and buttons are still in the dom. 
This is obviously not acceptable as there is a big grey space between the widgets and the top of the page, of which, users can still interact with. 

1. Can you please fix this with I think one of two options:
   1. by using a slide up for hide, and slide down for show animation instead of the opacity. As we want the kpis-bar empty space to reduce as if the kpis-bar was not there,
   2. or, after the opacity finishes fading, then using transition to reduce / expand the height of the kpis bar. (if you use both opacity and height transition, increase the speed of the opacity fade on the kpis).

2. I think you should use a JS method for more control, remember you can apply ngShow/ngHide with the css class `'hide'` & `'show'`. This removes it from the DOM. So maybe you could do something like `angular.element(el).animate(...)` then  `$timout(-> element.addClass('hide'), durationOfSlideAnimation`.

Take a look at kpi.directive.coffee `animateKpiPanels` method for inspiration.

Make sure when in hidden mode, the KPIs and kpis-bar buttons are not visible or interactable.

xaun:
Merged into https://github.com/maestrano/impac-angular/pull/257

PR #254, opened 2016-09-19
Fix sync statuses modal display

cesar-tonnoir:
@xaun can you review please?


cesar-tonnoir:
@xaun it looks like your commit #e8d999b is not in upstream yet?

PR #255, opened 2016-09-21
[IMPAC-343] Save kpis on enter keydown

an-pavlovych:


an-pavlovych:
Maybe also necessary to switch `editMode`?

xaun:
Merged here https://github.com/maestrano/impac-angular/pull/258

PR #262, opened 2016-10-21
Release/1.4.4

xaun:
@cesar-tonnoir please review

### DIFF
--- a/src/components/dashboard-settings/sync-apps/sync-apps.directive.js.coffee
+++ b/src/components/dashboard-settings/sync-apps/sync-apps.directive.js.coffee
@@ -87,7 +87,7 @@ module.directive('dashboardSettingSyncApps', ($templateCache, $log, $http, $filt
         $http.post(ImpacRoutes.organizations.appInstancesSync(scope.orgUID)).then( (resp) -> processAppInstancesSync(resp.data) )
 
       scope.triggerSyncAlertsModal = ->
-        unless scope.modalOpened || !(scope.connectors && scope.connectors.length)
+        unless scope.modalOpened || _.isEmpty(scope.connectors)
           modalInstance = $modal.open({
             animation: true
             size: 'md'
--- a/src/components/dashboard-settings/sync-apps/sync-apps.directive.js.coffee
+++ b/src/components/dashboard-settings/sync-apps/sync-apps.directive.js.coffee
@@ -87,7 +87,7 @@ module.directive('dashboardSettingSyncApps', ($templateCache, $log, $http, $filt
         $http.post(ImpacRoutes.organizations.appInstancesSync(scope.orgUID)).then( (resp) -> processAppInstancesSync(resp.data) )
 
       scope.triggerSyncAlertsModal = ->
-        unless scope.modalOpened
+        unless scope.modalOpened || !(scope.connectors && scope.connectors.length)
           modalInstance = $modal.open({
             animation: true
             size: 'md'
--- a/src/components/widget/widget.directive.coffee
+++ b/src/components/widget/widget.directive.coffee
@@ -60,6 +60,7 @@ module.controller('ImpacWidgetCtrl', ($scope, $log, $q, $timeout, ImpacWidgetsSv
     $scope.pdfMode = true
     w.ticked = true unless angular.isDefined(w.ticked)
     $scope.widget.hasEditAbility = false
+    $scope.initSettings()
   )
   ImpacDashboardsSvc.pdfModeCanceled().then(null, null, ->
     $scope.pdfMode = false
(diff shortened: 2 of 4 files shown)


src/components/dashboard-settings/sync-apps/sync-apps.directive.js.coffee
```diff
@@ -87,7 +87,7 @@ module.directive('dashboardSettingSyncApps', ($templateCache, $log, $http, $filt
         $http.post(ImpacRoutes.organizations.appInstancesSync(scope.orgUID)).then( (resp) -> processAppInstancesSync(resp.data) )
 
       scope.triggerSyncAlertsModal = ->
-        unless scope.modalOpened
+        unless scope.modalOpened || !(scope.connectors && scope.connectors.length)
```
cesar-tonnoir: `!_.isEmpty(scope.connectors)` ?

src/components/widgets-settings/width/width.directive.coffee
```diff
@@ -62,14 +62,13 @@ module.controller('SettingWidthCtrl', ($scope, $element, $timeout, $log, ImpacWi
...
     $scope.pdfMode = true
-    $scope.initiallyExpanded = $scope.expanded
     # Expand the widget if it's not already the case
     w.toggleExpanded(false) unless $scope.expanded
   )
   ImpacDashboardsSvc.pdfModeCanceled().then(null, null, ->
     $scope.pdfMode = false
     # Reduce the widget if it wasn't expanded initially
-    w.toggleExpanded(false) unless $scope.initiallyExpanded
+    w.toggleExpanded(false) unless w.initiallyExpanded
```
cesar-tonnoir: Not a good practice there: `w.initiallyExpanded` is initialized in another scope (`DashboardSvc`) => quite hard to maintain...
Can't we rather pass an attribute to the directive (`scope.expanded`)?...

xaun: @cesar-tonnoir what other scope is `w.initiallyExpanded` initialised in? Im not a big fan of these variables being attached to the widget either.. But seeing as `w.initialWidth`, and `w.width` are being initialised in the widget.directive, it makes sense to me to also initialise an `w.initiallyExpended` value. If we do pass an attribute, we will need to add that to every widget template `<settings-width>` attributes.

cesar-tonnoir: my mistake: it's not initialized in DashboardSvc, but in widget.directive. I'll describe in another comment

src/components/widget/widget.directive.coffee
```diff
@@ -53,19 +53,22 @@ module.controller('ImpacWidgetCtrl', ($scope, $log, $q, $timeout, ImpacWidgetsSv
     return classes.join(' ')
 
   ImpacDashboardsSvc.pdfModeEnabled().then(null, null, ->
-    # if w.expanded is defined, the width directive will handle the display
-    unless angular.isDefined(w.expanded)
+    # Save whether the widget is expanded
```
cesar-tonnoir: Shouldn't we should move the logic related to the width management in the width directive?
Maybe something like:
```coffeescript
# in widget directive
ImpacDashboardsSvc.pdfModeEnabled().then(null, null, ->
  $scope.pdfMode = true
  w.ticked = true unless angular.isDefined(w.ticked)
  $scope.widget.hasEditAbility = false

# in widget directive
ImpacDashboardsSvc.pdfModeCanceled().then(null, null, ->
  $scope.pdfMode = false
  $scope.widget.hasEditAbility = true

# ----------------------------------

# in width directive
ImpacDashboardsSvc.pdfModeEnabled().then(null, null, ->
  $scope.pdfMode = true
  if w.isExpanded()      # should always be defined in this case
    $scope.initiallyExpanded = true
  else
    $scope.initialWidth = w.width
    w.width = 12
    # I don't think the below will still be necessary?
    # $scope.initSettings()
  w.toggleExpanded(false) unless $scope.expanded

# in width directive
ImpacDashboardsSvc.pdfModeCanceled().then(null, null, ->
  $scope.pdfMode = false
  unless w.isExpanded()      # should always be defined in this case
    w.width = $scope.initialWidth
  w.toggleExpanded(false) unless $scope.initiallyExpanded
```

Imho it would be more clear as the concerns would be separated between what regards the width and the widget in itself, + you would minimise the use of variables attached to `w`...

By looking at it closer, I realise there is something very weird with the width management: have a look at the `toggleExpanded` function (in width directive): it looks like we are modifying the width again there... You can probably also wrap some code of the sample above in the `toggleExpanded` function and reduce the number of variables / methods used...

What do you think?

cesar-tonnoir:
Changes approved :+1:

PR #264, opened 2016-11-17
Internationalisation of Impac angular

clemthenem:
This is the work I have done so far for the internationalization of impac-angular.


cesar-tonnoir:
Ok with the general principles. Each file must be tested one by one when we merge the feature branch to a version branch.

PR #269, opened 2016-11-30
Added the missing words to the en-gb dictionary

Can9UA:


xaun:
Hi @Can9UA, just trying to get your branch running in my local so I can review. Having some problems getting it running. There seems to be a lot of interpolation errors?

screen shot 2016-12-08 at 15 51 18

Any ideas?

src/components/dashboard/dashboard.tmpl.html
```diff
@@ -110,8 +110,8 @@ <h3 translate>impac.dashboard.it_looks_like</h3>
           <div class="row top-buffer-4">
             <div class="col-md-8 col-md-offset-2">
               <div class="info-text">
-                <p><b>{{dhbErrorsConfig.firstTimeCreated.first}}</b></p>
-                <p>{{dhbErrorsConfig.firstTimeCreated.second}}</p>
+                <p><b translate>{{dhbErrorsConfig.firstTimeCreated.first}}</b></p>
```
Can9UA: If 'dhbErrorsConfig.firstTimeCreated.first' returns a string that does not match a translation variable name in dictionary, translation function will return the string which will be equal variable name.

For example:
<p translate>{{'name of a non-existent variable'}}</p>
will be:
<p>name of a non-existent variable</p>

src/components/dashboard/dashboard.tmpl.html
```diff
@@ -27,10 +27,9 @@
 
     <!-- Displayed to MYOB Essentials Users -->
     <div id="myob-message" ng-if="hasMyobEssentialsOnly && myobMessageConfig.show" class="alert alert-warning">
-      <h3>It looks like you are using MYOB Essentials...</h3>
-      <p>Unfortunately, all the information required by Impac! is not exposed by MYOB Essentials API <strong>yet</strong>.
-      Consequently, most of the accounting widgets provided by Impac! won't find a valid source of data to display their content.</p>
-      <p>Maybe you should consider switching to <a href="http://myob.com.au/products/small-business/accounting-software/myob-accountright-1422735752948" target="_blank">MYOB AccountRight</a>, the other MYOB accounting solution?</p>
+      <h3 translate>impac.dashboard.it_looks_like</h3>
+      <p translate="impac.dashboard.unfortunately_information"></p>
```
Can9UA: Yes, it is correct. It is for compiling string to html if this string is contain some html tags. In this case: 
"impac.dashboard.unfortunately_information": "Unfortunately, all the information required by Impac! is not exposed by MYOB Essentials API <strong>yet</strong>......."

src/components/widgets/accounts-class-comparison/accounts-class-comparison.directive.coffee
```diff
@@ -23,14 +23,21 @@ module.controller('WidgetAccountsClassComparisonCtrl', ($scope, $q, $filter, Cha
 
   # Widget specific methods
   # --------------------------------------
+  
+  translate = (word) ->
+    word = word.toLowerCase()
+    translation = $translate.instant('impac.widget.account_class_comp.klass.' + word)
+    return if ~translation.indexOf('impac.widget.account_class_comp.klass') then _.capitalize(word) else translation
```
xaun: Please use lodash here instead of bitwise operator

PR #270, opened 2016-11-30
Refactor routes service & update workspace config

xaun:
@cesar-tonnoir please review

### DIFF
--- a/src/services/routes/routes.svc.coffee
+++ b/src/services/routes/routes.svc.coffee
@@ -1,5 +1,4 @@
 # provider for configuring http endpoints.
-# TODO: REFACTOR THIS FOR MNOE AS DEFAULT + REVISE BASE PATHS DEFAULTS.
 angular
   .module('impac.services.routes', [])
   .provider('ImpacRoutes', () ->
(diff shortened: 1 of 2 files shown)


src/services/routes/routes.svc.coffee
```diff
@@ -9,7 +8,7 @@ angular
     #=======================================
     defaults =
       # base paths
-      mnoHub: '/api/v2'
+      mnoHub: 'http://localhost:7000/mnoe/jpi/v1'
       impacPrefix: '/impac'
```
cesar-tonnoir: Do we still have a case where `impacPrefix` wouldn't be `/impac` ? I think this was done for Maestrano and `/analytics` ?...
If the case does no longer exist I suggest to simply remove the variable or to add it to the `mnoHub` path variable

src/services/routes/routes.svc.coffee
```diff
@@ -99,11 +92,8 @@ angular
           if defaults.widgets.index
             defaults.widgets.index.replace(':dashboard_id', dashboard_id)
           else
-            "#{service.dashboards.show(dashboard_id)}/widgets"
+            "#{defaults.mnoHub}#{defaults.impacPrefix}/widgets"
```
cesar-tonnoir: shallow routes FTW! :)

xaun:
Hey @cesar-tonnoir, regarding your question, express projects are by default manually configuring impac-angular's routes for mnoe, so this change will not impact the express projects. Each express project's impac.config can be refactored to use the defaults at any stage.  Let me know how this should be released, I think just an email to the mnoe team letting them know the projects can be refactored should suffice? 

I have pushed a commit for the removal of the `impacPrefix`, please review. :beers:

PR #275, opened 2016-12-13
[IMPAC-342] data not found display, refactors, & bug fixes.

xaun:
@cesar-tonnoir please review. 

@hedudelgado also please review, especially the "Fix data not found & kpi loader display logic" commit for our pair programming session.

@MarcoCode thanks for your eyes on commits "Fix available kpis logic & tenant disabled display" & "Addd guard for kpi.directive init extra_params for no accounts"


src/components/kpi/kpi.directive.coffee
```diff
@@ -112,8 +112,17 @@ angular
           ImpacKpisSvc.validateKpiTargets($scope.targets)
 
         $scope.hasContent = ->
-          return true if $scope.isEditing()
-          $scope.kpi && $scope.kpi.layout && $scope.kpi.data
+          !!($scope.kpi && $scope.kpi.layout && $scope.kpi.data)
+
+        $scope.showKpiContent = ->
+          # Newly added kpis start in edit mode (draft) and will have "no content".
+          !$scope.isLoading() && ($scope.hasContent() || $scope.kpi.isDraft)
```
cesar-tonnoir: `!$scope.isLoading() && !$scope.isDataNotFound()`

src/components/common/data-not-found/data-not-found.directive.coffee
```diff
@@ -0,0 +1,63 @@
+module = angular.module('impac.components.common.data-not-found',[])
+
+module.directive('commonDataNotFound', ($templateCache, $log, $http, ImpacAssets, ImpacTheming, ImpacMainSvc) ->
+  return {
+    restrict: 'A',
+    scope: {
+      isKpi: '=?'
+      engine: '='
```
cesar-tonnoir: I would rename that to `endpoint` instead to match `kpi.endpoint` (+ at some point `widget_category` will be renamed to `endpoint` as well...)

src/components/common/data-not-found/data-not-found.directive.coffee
```diff
@@ -0,0 +1,63 @@
+module = angular.module('impac.components.common.data-not-found',[])
+
+module.directive('commonDataNotFound', ($templateCache, $log, $http, ImpacAssets, ImpacTheming, ImpacMainSvc) ->
+  return {
+    restrict: 'A',
+    scope: {
+      isKpi: '=?'
```
cesar-tonnoir: instead of `isKpi`, I would recommend to propose a more general parameter like for example `hasImage` => this way you can use the directive for any component which can be a widget, a kpi, or anything else...

src/components/common/data-not-found/data-not-found.directive.coffee
```diff
@@ -0,0 +1,63 @@
...
+    link: (scope, element) ->
+      scope.content     = ImpacTheming.get().dataNotFoundConfig
+      widgetImgsBaseDir = ImpacAssets.get('dataNotFound')
+      usingDefaults     = false
+      imagePath         = ''
+      image             = element.find('.data-not-found #not-found-bg')[0]
+
+      image.onerror = ->
+        missingImageLocationMsg = if usingDefaults then 'library defaults' else scope.engine
+        $log.warn("Missing data-not-found image for #{missingImageLocationMsg}")
```
cesar-tonnoir: Is it going to be executed for KPIs ?

src/components/kpi/kpi.tmpl.html
```diff
@@ -61,8 +61,16 @@
 
       </div>
     </div>
+    <!-- Data not found display -->
+    <div class="kpi-no-data" ng-if="isDataNotFound()">
+      <!-- Remove KPI -->
+      <button type="button" ng-show="isEditing()" class="kpi-remove im-fab im-warn" ng-click="deleteKpi()">
+        <i class="fa fa-times" aria-hidden="true"></i>
+      </button>
+      <div common-data-not-found is-kpi="true" widget-engine="::kpi.endpoint"/>
```
cesar-tonnoir: warning there I think you've renamed `widget-engine` in the directive. Don't you have to update widget.tmpl.html as well ?

src/components/common/data-not-found/data-not-found.directive.coffee
```diff
@@ -0,0 +1,63 @@
...
+          return loadCustomWidgetImage() if widgetImgsBaseDirProvided()
+          scope.hasCallbackUrl = scope.content.linkUrlCallback?
+          # When providing no custom images, uses library defaults.
+          displayDefaultImage()
+
+      load()
+
+    template: $templateCache.get('common/data-not-found.tmpl.html'),
+  }
+)
```
cesar-tonnoir: By reading this file, I have the impression that the whole logic is by-passed when used on a kpi, or am I missing something?
If this is the case, why reusing this directive instead of simply applying the "data-not-found" style to a div with a message? I mean, if we don't need the logic that fetchs the image and loads it in the background, the need for a directive is not that obvious?...

src/components/common/data-not-found/data-not-found.directive.coffee
xaun: No, first I did extend the image capabilities for KPIs. But then I realised that an image background will not fit / look good with the KPIs. So I removed the ability. So indeed the image logic is bypassed for kpis.

src/components/kpi/kpi.tmpl.html
xaun: Ah you are right, I have messed this up. I will fix this by renaming it to `endpoint` as suggested in your comment above.

src/components/common/data-not-found/data-not-found.directive.coffee
xaun: Yes I agree with this. It was isKpi as kpi image capabilities where developed at one point in the history of this branch.

src/components/kpi/kpi.directive.coffee
xaun: The logic in `isDataNotFound` is slightly different... using an `&&` rather than a `||`. 
```
$scope.isDataNotFound = ->
          !$scope.hasContent() && !$scope.kpi.isDraft
```

src/components/common/data-not-found/data-not-found.directive.coffee
cesar-tonnoir: Ok, so read my message again: I did not recommend to duplicate the directive template, I was just challenging the fact that maybe we don't need a full directive for just a div with a fixed string inside.
In any case, I was asking questions, not recommending or enforcing anything in particular. The way it's done works and is DRY => ok for me.

PR #292, opened 2017-01-17
[IMPAC-335] Multiple Recipients 

adamaziz15:


xaun:
https://github.com/maestrano/mno-enterprise-angular/pull/125

adamaziz15:
@xaun @cesar-tonnoir I have begun modifying the frontend to send up multiple recipients when creating an alert. My first concern was to make sure the data was going to mno-hub properly so that it could be used to create the alert. This is where I got a little stuck, I think her is removing the data from params. I am not too familiar with it so I am still figuring things out. 😅

src/services/kpis/kpis.svc.coffee
```diff
@@ -385,7 +385,7 @@ angular
       promises = []
 
       for alert in alertsToCreate
-        promises.push ImpacAlerts.create(kpi.id, { alert: _.pick(alert, ['service']) })
+        promises.push ImpacAlerts.create(kpi.id, { alert: _.pick(alert, ['service']), recipients: alert.recipients })
```
xaun: @adamaziz15 an in-app alert wont have key `recipients`, maybe do a merge if service is email instead.

adamaziz15:
@xaun I've improved the UI to initialize org members as active or inactive so that in the alerts modal it will show who is already receiving the alert. I've also added the appropriate functions to all of the services needed to allow for updating alerts.

adamaziz15:
@cesar-tonnoir @xaun Should be ready to go 👍 , Please Review.

src/components/alerts-config/alerts-config.modal.html
```diff
@@ -9,12 +9,36 @@
...
-        {{alert.label}}
+      <li class="list-group-item" ng-repeat="alert in alerts" ng-class="{active: alert.active}">
+        <div ng-click="toggleAlert(alert)">
+          <i class="fa" ng-class="alert.active ? 'fa-toggle-on' : 'fa-toggle-off'" />
+          {{alert.label}}
+        </div>
+
+        <ul class="list-group recipients-list" ng-show="showRecipientList(alert)">
+
+          <li class="list-group-item" ng-repeat="member in members" ng-show="member.active" ng-click="toggleRecipient(member)">
```
xaun: I think `ng-if` should be used instead of `ng-show` here. The difference is `ng-if` will remove the element from the dom, `ng-show` will just apply/remove a class with visibility styles.

src/components/alerts-config/alerts-config.directive.coffee
```diff
@@ -64,5 +83,16 @@ module.directive('alertsConfig', ($modal, $templateCache, $compile, ImpacKpisSvc
           $scope.alerts[alert.service].active = true
 
         $scope.modal = $modal.open(alertsSettingsModal.options)
+
+      $scope.showRecipientList = (alert) ->
+        alert.active && alert.service == 'email'
+
+      $scope.onKeyPress = (event) ->
+        if event.which == 13
+          recipient = _.filter($scope.members, (member) -> member.email == $scope.search.text)[0]
```
xaun: `_.find()` instead

src/components/alerts-config/alerts-config.modal.html
```diff
@@ -9,12 +9,36 @@
...
+            {{member.email}}
+            <button type="button" class="close" aria-label="Close">
+              <span aria-hidden="true">&times;</span>
+            </button>
```
xaun: Doesn't look like this remove button is hooked up to a method?

src/components/alerts-config/alerts-config.modal.html
```diff
@@ -9,12 +9,36 @@
...
+          </li>
+
+        </ul>
+
+        <input class="recipient-search" placeholder="Email" ng-show="showRecipientList(alert)" ng-model="search.text" ng-keypress="onKeyPress($event)"/>
+
+        <ul class="list-group recipients-list available" ng-show="showRecipientList(alert)">
+
+          <li class="list-group-item" ng-repeat="member in members | filter:search.text" ng-hide="member.active" ng-click="toggleRecipient(member)">
```
xaun: Like above, `ng-if="!member.active"` to remove element from dom.

src/services/kpis/kpis.svc.coffee
```diff
@@ -382,14 +382,33 @@ angular
         !alerts[alert.service].active
       )
 
+      # Update existing alerts that have been modified
+      alertsToUpdate = _.filter(alerts, (alert) ->
+        active_alert = _.filter(kpi.alerts, (kpi_alert) -> kpi_alert.service == alert.service)[0]
```
xaun: Rename to `existingAlert`and use `_.find()`

src/services/kpis/kpis.svc.coffee
```diff
@@ -382,14 +382,33 @@ angular
...
+        return false if alert.service == "inapp" || !active_alert
+        active_ids = active_alert.recipients.map((recipient) -> recipient.id).sort() if active_alert
+        recipient_ids = alert.recipient_ids.sort()
+        recipient_change = false
+        recipient_change = true if recipient_ids.length != active_ids.length
+        for i in [0...recipient_ids.length]
+          recipient_change = true if recipient_ids[i] != active_ids[i]
```
xaun: Lets simplify the change detection above by comparing the arrays like so: `"#{[1, 3, 2].sort()}" == "#{[2, 1, 3].sort()}"`, also I think it should be clear with the variable naming what is existing and what is new. So naming `active_ids` to `existingActiveIds` etc.

src/components/alerts-config/alerts-config.modal.html
adamaziz15: Because the ng-click directive is on the <li> element it is triggered when the button is clicked. The button is really there more for UX purposes as it makes it more clear which members are receiving the alert as well as how to remove them.

cesar-tonnoir:
@xaun / @adamaziz15 From what I've seen so far I think this won't be a problem, but please verify that there is no backward incompatibility with mno-enterprise (= if you upgrade impac-angular without upgrading mno-enterprise, you can still add alerts with recipients)

adamaziz15:
@cesar-tonnoir @xaun I made a few changes to make sure there is no issue with backward-compatibility and tested everything. Seems to work with no issues. Please review, should be all set :ok_hand:

PR #293, opened 2017-01-18
[IMPAC-440] fix sorting bugs

xaun:


xaun:
Impac! PR: https://github.com/maestrano/impac/pull/421

PR #295, opened 2017-01-19
[IMPAC-463] No KPIs data when changing to multi-organisation w/ shared dhb

xaun:
@cesar-tonnoir when KPIs array is the same (has same id's) for a different organization's dashboard (multiple data source dashboard), `track by kpi.id` prevents KPI directive
controllers from detecting change and reloading.. Took me a while to find this difference between the widgets and the KPIs load architecture! Please review :beers:


cesar-tonnoir:
fix conflict on Changelog.md

PR #299, opened 2017-01-25
[IMPAC-346] Add KPIs bar information section

hedudelgado:
nohover
hover
information
informationscrollable
settings

@cesar-tonnoir  @xaun


src/components/kpi/kpi.less
```diff
@@ -206,6 +235,7 @@
     background: @kpi-settings-remove-background;
     &:hover {
       background: @kpi-settings-remove-background-hover;
+      background-color: @mblue;
```
xaun: Can we change this to be the same colour as the kpi-bar action buttons on hover

src/components/kpi/kpi.tmpl.html
```diff
@@ -13,6 +14,14 @@
             <strong>{{kpi.layout.text.emphasis}}</strong> - <span class="real-value">({{ getRealValue() }})</span>
           </span>
         </div>
+        <!-- Info KPI -->
+        <div class="kpi-edit">
```
xaun: As mentioned, remove the container with the kpi-edit class. And move the `.impac-material(@kpi-settings-focus-color, @kpi-settings-error-color);` so it is applied to anything within the `.kpi` class.

PR #300, opened 2017-01-31
[IMPAC-465] Fix dates picker positioning

xaun:
Bootstrap-ui v2.x uib-datepicker-popup compiles a wrapping directive
("uib-datepicker-popover-wrap") in the same scope level for encapsulating
the "popped up" bootstrap dates picker template (on-click the datepicker is
rendered within it). This element is `display: block`, which causes issues when
elements along side are `display: inline-block`..

Wrapping the buttons in an inline-block styled div means that the uib-datepicker-popover-wrap directive is compiled within that (along side the button), and everything is happy.

This issue only affects impac-angular versions >= 1.5.0rc1, as bootstrap handles this differently (using a compiled `<li>`) in the version used in impac-angular 1.4.x.

Btw, mnoe-demo has commit history saying it's pointing to impac-angular#1.5.0rc2, but it's built with 1.5.0-rc1.

@alexnoox have you seen this? 

@cesar-tonnoir please review


alexnoox:
No, i've not seen this. If I understand well what you say, when you open the popover, the layout is acting weird?

In the doc, there is a `datepicker-append-to-body` property, have you tried it?

alexnoox:
@cesar-tonnoir This is what makes the beauty of this profession.

xaun:
hahahaha, @alexnoox please write a book of original proverbs. I will buy it.

PR #301, opened 2017-02-01
[IMPAC-99] Widgets discovery

cesar-tonnoir:
- depends on https://github.com/maestrano/impac/pull/424
- depends on https://github.com/maestrano/mno-enterprise/pull/216
- backward-compatible with mnohub

### DIFF
--- a/src/services/dashboards/dashboards.svc.coffee
+++ b/src/services/dashboards/dashboards.svc.coffee
@@ -126,7 +126,14 @@ angular
               )
 
               # Retrieve widgets templates
-              $http.get(ImpacRoutes.widgets.templates()).then(
+              url = 
+                if ImpacTheming.get().dhbWidgetsConfig.templates.defaultToFinancialYear
+                  fy_end_month = ImpacMainSvc.getFinancialYearEndMonth()
+                  "#{ImpacRoutes.widgets.templates()}?financial_year_end_month=#{fy_end_month}"
+                else
+                  ImpacRoutes.widgets.templates()
+
+              $http.get(url).then(
                 (response) ->
                   _self.setWidgetsTemplates(response.data.widgets)
                   _self.loadLocked = false
--- a/src/services/dashboards/dashboards.svc.coffee
+++ b/src/services/dashboards/dashboards.svc.coffee
@@ -1,6 +1,6 @@
 angular
   .module('impac.services.dashboards', [])
-  .service 'ImpacDashboardsSvc', ($q, $http, $log, $timeout, ImpacMainSvc, ImpacRoutes, ImpacTheming, ImpacDeveloper) ->
+  .service 'ImpacDashboardsSvc', ($q, $http, $log, $timeout, ImpacMainSvc, ImpacRoutes, ImpacTheming, ImpacDeveloper, ImpacUtilities) ->
     #====================================
     # Initialization and getters
     #====================================
@@ -206,7 +206,6 @@ angular
           $log.error("Impac! - DashboardsSvc: Cannot load user's organizations")
       )
 
-
     @setWidgetsTemplates = (srcTemplates) ->
       return false if _.isEmpty(srcTemplates)
       srcTemplates = ImpacDeveloper.stubWidgetsTemplates(srcTemplates) if ImpacDeveloper.isEnabled()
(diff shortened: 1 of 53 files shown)


cesar-tonnoir:
@xaun I think this one is ready too.
Quite important changes - this will depend on impac#1.5.0 and mnoe-enterprise#3.3.0
I've also done a bit of styling when I had the occasion...
Ask if any question.

I will start the testing as soon as I have your 'ok'. This should be part of the 1.5.0 release of next week, with internationalisation and angular 1.6 upgrade

src/services/developer/developer.svc.coffee
```diff
@@ -27,14 +27,7 @@ angular
...
     _$get = ($q) ->
-      # Mapping constant for widget stubbing.
-      # TODO: impac-angular receives a template with `template.path`, converts
-      #       it to a `widget.widget_category` key, then server responds with
-      #       `widget.category`. This should be more consistant.
-      WIDGET = {
-        widget_category: 'path'
-        category: 'path'
-      }
+      WIDGET = { endpoint: 'endpoint' }
```
xaun: Wohoo 👍 🍾

src/services/developer/developer.svc.coffee
```diff
@@ -27,14 +27,7 @@ angular
```
xaun: Have you tested that the widget stubbing works with the dev toolkit? Can test really quickly with the toolkit settings page. Looks good just wondering if you've tried it

src/services/dashboards/dashboards.svc.coffee
```diff
@@ -110,26 +110,47 @@ angular
...
+              # Retrieve dashboards
               $http.get(ImpacRoutes.dashboards.index(orgId)).then(
                 (dashboards)->
                   _self.setDashboards(dashboards.data).then(->
                     _self.setCurrentDashboard()
                     deferred.resolve(_self.config)
                     $log.info("Impac! - DashboardsSvc: loaded (force=#{force})")
 
-                  ).finally( -> _self.loadLocked=false )
+                  ).finally( -> _self.loadLocked = false )
```
xaun: I think this finally may resolve before the retrieval of templates below is finished, unlocking the `loadLocked`.

src/services/dashboards/dashboards.svc.coffee
```diff
@@ -110,26 +110,47 @@ angular
             (success)->
               orgId = success.currentOrganization.id
 
+              # Retrieve dashboards
               $http.get(ImpacRoutes.dashboards.index(orgId)).then(
                 (dashboards)->
                   _self.setDashboards(dashboards.data).then(->
                     _self.setCurrentDashboard()
                     deferred.resolve(_self.config)
                     $log.info("Impac! - DashboardsSvc: loaded (force=#{force})")
```
xaun: DashboardSvc may not really be loaded here. This could log as successful, but the widget retrieval below could fail.

src/services/dashboards/dashboards.svc.coffee
```diff
@@ -110,26 +110,47 @@ angular
...
+                  fy_end_month = ImpacMainSvc.getFinancialYearEndMonth()
+                  "#{ImpacRoutes.widgets.templates()}?financial_year_end_month=#{fy_end_month}"
+                else
+                  ImpacRoutes.widgets.templates()
+
+              $http.get(url).then(
+                (response) ->
+                  _self.setWidgetsTemplates(response.data.widgets)
+                  _self.loadLocked = false
+
```
xaun: Hmm I'm really not sure if this method will run consistently. `@load` returns a deferred, so one would expect this success handler here to `deferred.resolve()` something... But the get dashboards call above success handler has a deferred resolve. Not really a good idea to have two success handlers like this, the code invoking this function will get some pretty weird responses should it define success / error handlers. 

Maybe you should consider using $q.all ? Or moving this into another method that gets called on dashboards load success?

src/services/developer/developer.svc.coffee
cesar-tonnoir: lol ^^ Impac-99 mate... it's been a while...

src/services/developer/developer.svc.coffee
cesar-tonnoir: It's working even better than before: now if you try to add a template in the workspace without having defined a proper component related to it, it won't break ;)

src/services/dashboards/dashboards.svc.coffee
cesar-tonnoir: Very good point. I've refactored it to use $q.all and added a few comments.
Url is also now retrieved using a different function for more clarity.
dashboard-loading
On the screenshot above, we can see that `/widgets` and `/dashboards` are called at the same time, and the widgets and kpis start loading only after the two calls return.

Btw... `/widgets` is 11 times faster than `/dashboard` through `Her`... :)

PR #302, opened 2017-02-06
[IMPAC-331] Display layout & calculations on first add

adamaziz15:
@xaun @cesar-tonnoir This PR should be ready to go. KPI's now have context on first add 👍. Cheers!


src/components/kpi/kpi.less
```diff
@@ -154,13 +154,30 @@
...
+      background: lighten(@kpi-negative-alert-label-background-color, 20%);
+      border-radius: 3px;
+      width: 100%;
       margin: 0 auto 5px auto;
       box-shadow: none;
       padding: 2px;
+      color: white;
+    }
+
+    .im-button.alert-config-btn {
```
cesar-tonnoir: not too sure about this... have you checked it's still working on mobile?
I let @xaun judge for the color and layout

Generally speaking I would keep this kind of change in a separate pull request as it can be subject to discussion and block the other fixes your PR is aiming to resolve...

src/components/kpi/kpi.tmpl.html
```diff
@@ -20,11 +20,15 @@
           <h6>{{kpi.name.toUpperCase()}}</h6>
         </div>
         <alerts-config kpi="kpi" button-html="
-          <div class='im-button alert-caption'>
-            <span><a href='' ng-bind='kpi.layout.text.alert' /></span>
+          <div class='im-button alert-config-btn'>
+            <i class='fa fa-bell'></i>
```
cesar-tonnoir: If this 'bell' button appears on the top of the kpi, I reckon it should be placed somewhere else, close to the "delete" button for example (instead of using 'absolute' + negative margin)

src/components/kpi/kpi.less
```diff
@@ -154,13 +154,30 @@
       text-align: center;
     }
 
-    .im-button.alert-caption {
+    .alert-caption {
       text-align: center;
       font-size: 11px;
-      width: 81%;
+      background: lighten(@kpi-negative-alert-label-background-color, 20%);
```
xaun: I don't think we should use lighten here, when express projects are briefed to use a certain colour hexcode, it will cause problems :) You can use lighten in the variables.less to help you define the impac-angular default.

src/components/kpi/kpi.tmpl.html
xaun: Yes, @adamaziz15 please move this down in the HTML so it is above the kpi remove button & commented. The styling is absolute so it wont effect the positioning, but the html will read better.

src/components/kpi/kpi.less
xaun: @adamaziz15 the `.im-button` class is for creating a impac-material square button. For the rounded buttons, you can utilise `.im-fab` (which will take care of the rounding you're doing). Take example from the remove button. The alert button should be `<button class="kpi-alerts im-fab im-primary">`. Then add your custom less styles above the kpi-remove button's styles in this file, customising the positioning. Also, I think the button style theme is to be black, and then the colour (blue in this instance) on hover. If implemented alike the remove button, it should work on mobile.

adamaziz15:
@xaun @cesar-tonnoir Made all of the recommended changes and added another screenshot to JIRA, please review :)

xaun:
@cesar-tonnoir pending your review on this one, I'll manage rebasing it onto a new release branch.

PR #309, opened 2017-03-07
[IMPAC-506] - Added currency-rates service

fgourichon:
@cesar-tonnoir for review


cesar-tonnoir:
Code looks good to me. How hard would it be to query the Connec! report instead of Impac!?

PR #317, opened 2017-04-03
[IMPAC-498] widget improve drill-down saved ids

xaun:
@cesar-tonnoir please review.

I took a look at trying to refactor these settings out into directives, but they're all a bit different in one way or another and I'm not sure its worth the time right to try streamline these right now?

Secondly, take a look at the payroll_summary.rb engine, it is building an id similar to what I have done in the frontend for the other widgets. Im wondering whether the api should do similar for all widgets? But I feel like it's not great to be doing that in Impac! because then in the future if we want to use the id for a post action of some-sort, we'd have to cleanse the id of the category prefix.


src/components/widgets/hr-payroll-summary/hr-payroll-summary.directive.coffee
```diff
@@ -146,6 +144,10 @@ module.controller('WidgetHrPayrollSummaryCtrl', ($scope, $q, ChartFormatterSvc,
 
   $scope.hasElements = ->
     $scope.selectedElements? && $scope.selectedElements.length > 0
+
+  getIdentifier = (element)->
```
cesar-tonnoir: do you think we could extract the method to ImpacUtilities ?

xaun: I thought about this but I don't think it belongs there, it is not generalised enough in my opinion. It should live within the setting component controller once refactored.

xaun:
@cesar-tonnoir I have pushed a refactor based on your suggestion - please review. 
I have decoupled the `selectedElement` from the `element` itself, leveraging `angular.copy`, to treat the `$scope.selectedElement` / `$scope.selectedElements` as a form of 'metadata'. Currently the only reason we need to attach a `category` attr onto the selected elements is because of the settings `toMetadata` method.. The `toMetadata` callback doesn't have access to the a `statementName` (any ideas here?). Im pretty happy with this refactor as the real collection of elements are left unmodified / unlinked, add syntax / performance improvements have been made.

Let me know, :beers:

PR #320, opened 2017-04-05
[IMPAC-527] sales segmented turnover minor display improvs

xaun:


xaun:
https://github.com/maestrano/impac/pull/450

PR #321, opened 2017-04-13
[IMPAC-537] Fix widget content css issues

xaun:


src/components/widget/widget.less
```diff
@@ -36,6 +36,7 @@
     border: @impac-widget-border;
     border-top: none;
     background-color: @impac-widget-background-color;
+    overflow: none;
```
cesar-tonnoir: @xaun I think you meant `hidden` ;)
I'll fix it.

PR #330, opened 2017-05-04
Improve widget selector using a Flexbox grid

xaun:
The main advantage is row columns will always be the same height, with a configurable min-height. 

Note: a Flexbox limitation is when a column's context wraps (e.g widget name), pushing beyond the minimum height, the columns on **that row** will match the height. But the columns on the previous/following wrapped rows will not ([See link here](http://stackoverflow.com/questions/36004926/is-it-possible-to-have-equal-height-columns-in-a-multi-line-flex-container)).

On the screenshot below, the screen size is reduce so some of the widget names are wrapping. The rows with at least one column with a wrapped widget name, are then all larger to accommodate this. As you may see, the last row's columns are smaller. This is apart of the limitation noted above. I still think this is a big win, and while not perfect looks natural.
screen shot 2017-05-04 at 14 28 00

The screenshot below has no widget names wrapping, so is flexing perfectly with a column height of 40px.
screen shot 2017-05-04 at 14 29 34

- Responsive behaviour tested on all browser screen sizes, ipad, iphone 5/6. 
- Browser compatibility tested on Chrome, Firefox & Safari.

FYI @ouranos @hedudelgado

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,6 +322,7 @@ Adjust Less styling by over-riding Less Variables in the parent application.
 // Widget selector
 @impac-dashboard-widget-selector-bg:            #233845;
 @impac-dashboard-widget-selector-text-color:    white;
+@impac-dashboard-widget-selector-widget-item-min-height: 55px;
 
 // Widgets container
 @impac-placeholder-border:                      2px dashed @impac-dashboard-borders-color;
--- a/src/components/dashboard/dashboard.tmpl.html
+++ b/src/components/dashboard/dashboard.tmpl.html
@@ -38,41 +38,41 @@
   <!-- Widgets selection container -->
   <div id="widget-selector" collapse="!showWidgetSelector()" ng-if="!customWidgetSelector.path">
     <div class="title">
-      <p>Select the widgets you want to add to your dashboard.</p>
-      <span class="badge confirmation">Widget added!</span>
-      <i class="fa fa-times-circle" ng-if="showCloseWidgetSelectorButton()" ng-click="displayWidgetSelector(false)"/>
+      <p class="instruction">Select the widgets you want to add to your dashboard.</p>
+      <span class="widget-added badge">Widget added!</span>
+      <i class="fa fa-times-circle close-selector" ng-if="showCloseWidgetSelectorButton()" ng-click="displayWidgetSelector(false)"/>
     </div>
 
     <div class="row top-container">
-      <div class="col-md-3 categories">
-        <div class="row header">
+      <div class="col-md-3 categories-section">
+        <div class="row section-header">
           All categories
         </div>
-        <div class="row lines">
-          <div class="col-md-12 categories_menu-items">
-            <p ng-click="selectCategory('accounts')" ng-class="isCategorySelected('accounts') ? 'selected' : none">Accounting</p>
-            <p ng-click="selectCategory('invoices')" ng-class="isCategorySelected('invoices') ? 'selected' : none">Invoicing</p>
-            <p ng-click="selectCategory('hr')" ng-class="isCategorySelected('hr') ? 'selected' : none">HR / Payroll</p>
-            <p ng-click="selectCategory('sales')" ng-class="isCategorySelected('sales') ? 'selected' : none">Sales</p>
+        <div class="row section-lines">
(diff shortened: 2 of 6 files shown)


cesar-tonnoir:
@xaun do you think you can make them all look like the screenshot 1? I mean: could they all be bigger regardless of the length of the name? I would like the display of the different categories to be consistent.
I'll review the PR in the meantime

README.md
```diff
@@ -322,6 +322,7 @@ Adjust Less styling by over-riding Less Variables in the parent application.
 // Widget selector
 @impac-dashboard-widget-selector-bg:            #233845;
 @impac-dashboard-widget-selector-text-color:    white;
+@impac-dashboard-widget-selector-widget-item-min-height: 40px;
```
cesar-tonnoir: Re my comment: I think the min-height could be adjusted to be high enough for 2 lines (maybe 60px?...)
Otherwise, good idea to put that in a variable.

src/components/dashboard/dashboard.tmpl.html
```diff
@@ -37,42 +37,42 @@
 
   <!-- Widgets selection container -->
   <div id="widget-selector" collapse="!showWidgetSelector()" ng-if="!customWidgetSelector.path">
-    <div class="title">
-      <p>Select the widgets you want to add to your dashboard.</p>
-      <span class="badge confirmation">Widget added!</span>
-      <i class="fa fa-times-circle" ng-if="showCloseWidgetSelectorButton()" ng-click="displayWidgetSelector(false)"/>
+    <div class="widget-selector__title">
```
cesar-tonnoir: Wouldn't it be preferable to keep the naming consistent with the rest of the lib?
```less
.widget-selector {
  .title {
    .instruction { }
  }
}
```

xaun: On this point, @cesar-tonnoir, and your message on slack "don't you think it will be a bit confusing for the developer if we have several conventions in the same lib?" I wanted to write down my thoughts on this and involve @alexnoox and start a little discussion, see what you guys think :)

In answer to your question, I feel like that’s kind of like saying refactoring some JS/Ruby code to follow a different convention/style from the rest of the project to improve maintainability & extensibility of that piece shouldn’t be done because it will confuse developers. 

I am just trying to experiment with some newer techniques, trying what people who put a lot of thought into making CSS easier to scale & more enjoyable to develop on are doing.

I do agree that applying a convention across a project is important, but I think applying new conventions to new / newly refactored components is a good way to make steps forward, making coming back to these components easier for our future selves.

So at this point you may be thinking what is the value of the BEM naming, and what is wrong with nesting in less/sass?

Most of our component classes are almost always nested 3+ levels deep, which creates some very specific classes. Specificity can, among other things:

- limit our ability to extend and manipulate stylesheets
- prevent things from working as expected when moved into different environments
- greatly reduce the modularity of our styles
- be very frustrating as a developer

My main goal with the way I want to be writing less/sass, is that I want my rules to match the elements I want, without them accidentally matching the elements I don’t, and where as the naming captures purpose and meaning that tells me more about what the piece of markup is doing.

So far trying out the BEM syntax I think one of the main advantages is that you have more idea of what the piece of markup is doing, and most importantly without the specificity. 

Originally when I looked at BEM I thought you had to represent each level on nesting with the double underscore. But let me show you an example:

What clicked for me is when I realised the 'block' is like defining the dependency for the 'element'. The `top-panel` element depends on the `widget` block, the `logo` element depends on the `top-panel`. So if I create another `widget` block, I know I can drop the `top-panel` and it's children classes in. 

```html
<div class="widget">
  <div class="widget__top-panel">
    <img class="top-panel__logo" />
    <p class="top-panel__text" />
  </div>
</div>
```

A bonus in being able to express a dependency on a modified block:

```html
<div class="section section--display-picture">
  ...
</div>
<div class="section section--about-me">
  <p class="section__body section--about-me__body" />
</div>
```

Once you experience making changes to classes like this, you realise that's its far less daunting to start moving things around, without the fear of breaking styles in unexpected locations. 

We don't have to follow the spec precisely, I think it's just about reducing Less/Sass nesting, and creating meaningful names that represent your design intentions.

PR #333, opened 2017-05-17
Set default row min height to 0px (widget selector)

cesar-tonnoir:
Tag not urgent - to be tagged with other fixes


cesar-tonnoir:
@xaun can you please validate if ok for you ?

xaun:
So what's up with this change? I thought you wanted to height to be the same so therefore larger?

PR #335, opened 2017-05-30
Bolt widgets integration

cesar-tonnoir:
- Routes to Bolts
- Cash Projection widget
- Cash Balance widget

**TODO**
- [x] Fix build
- [x] Refactor chart config to a service?


src/components/widgets/accounts-cash-balance/accounts-cash-balance.directive.coffee
```diff
@@ -0,0 +1,132 @@
+#
+# Component generated by Impac! Widget Generator!
+# TODO:
+# - colour shades by account sub-type (don't spend too much time on this one / not that important)
```
cesar-tonnoir: @xaun done, isn't it?

src/components/widgets/accounts-cash-balance/accounts-cash-balance.directive.coffee
```diff
@@ -0,0 +1,132 @@
...
+
+  setSeriesColors = (series, chartColors) ->
+    groupedSeries = _.groupBy(series, (serie)-> serie.bias)
+    for bias, series of groupedSeries
+      continue unless chartColors[bias]
+      palette = ImpacTheming.color.generateShadesPalette(chartColors[bias], series.length)
+      for serie, i in series
+        serie.color = palette[i]
+
+  buildChart = (data, onRender)->
```
cesar-tonnoir: @xaun
Can this part be simplified considering we are returning more display config from the API?
At least, maybe we could move that to a separate service so it can be re-used by several widgets?

src/components/widgets/accounts-cash-balance/accounts-cash-balance.directive.coffee
```diff
@@ -0,0 +1,132 @@
...
+
+  settingsPromises = [
+    $scope.orgDeferred.promise,
+    $scope.timePeriodDeferred.promise
+  ]
+
+  # Widget specific methods
+  # --------------------------------------
+  w.initContext = ->
+    if $scope.isDataFound = w.content?
```
cesar-tonnoir: For bolt widgets, `content` will always be defined - should we check if content.chart is defined instead?
TODO: when data not found, query the API again to load stub data.

src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
```diff
@@ -0,0 +1,75 @@
...
+    $scope.orgDeferred.promise,
+    $scope.timePeriodDeferred.promise
+  ]
+
+  # Widget specific methods
+  # --------------------------------------
+  w.initContext = ->
+    $scope.isDataFound = w.content?
+    
+    myChart = Highcharts.chart 'cashProjectionChart', {
```
cesar-tonnoir: Same thing here: maybe could it be refactored to a service?

src/filters/mno-currency/mno-currency.filter.coffee
```diff
@@ -4,7 +4,7 @@
 # $filter('mnoCurrency')(amount,currency,[true|false])
 #
 angular.module('impac.filters.mno-currency', []).filter('mnoCurrency', ($filter) ->
-  (amount, currency='', ISOmode=true) ->
+  (amount, currency='', ISOmode=true, decimal) ->
```
cesar-tonnoir: can you please document? where is that used?

src/components/widgets/accounts-cash-balance/accounts-cash-balance.directive.coffee
xaun: Well it depends, I think we should return `plotLines` from the API. But I think the highcharts config is very abstracted & is already very well wrapped. I don't see a huge advantage as most of the config here is widget specific. For example, the only re-usable config I see here is x & y axis `startOnTick`, `minPadding`, and `tickInterval`. On-top of this, there are [Themes](https://www.highcharts.com/docs/chart-design-and-style/themes), but not sure it's really worth it especially for this widget as it has a custom legend.

src/filters/mno-currency/mno-currency.filter.coffee
xaun: This gives the ability to set how many decimal places you would like. Before this change currency would always be 2 decimal places (which makes sense as a default), but the mockups had no decimal places. Where would you like this to be documented? Just a comment above?

PR #341, opened 2017-06-06
Improve tooltip display with date & currency

xaun:


xaun:
Replaced by https://github.com/maestrano/impac-angular/pull/342

PR #357, opened 2017-06-15
i18n - delegate locales loading to host app

cesar-tonnoir:
**I18n integration with mnoe is currently not working**
- The configuration of `$translateProvider` is conflicting with the config set in the host app
- The locales embedded in Impac! angular are preventing the host app to load its own

=> Instead of letting the host app configure impac! angular's language strategy, we would better expose the locale json files and let the host app configure the strategy for the full application.

**Proposed fix**
1. Impac! angular exposes its locale files
2. The host app configures $translateProvider
3. The host apps statically loads:
  a. its own locales
  b. impac's locales


cesar-tonnoir:
@ouranos @alexnoox here are the corresponding changes in the admin panel:
https://github.com/maestrano/mnoe-admin-panel/pull/25/commits/dd1565de1af372a8dc6be9b810df09e91c01e967

Note that there will be a dependency with MNOe as the host app will have to loads the locales exposed by Impac! (PR to follow).

workspace/app/app.js
```diff
@@ -132,4 +138,6 @@ module.run(function (ImpacLinking, ImpacAssets, ImpacRoutes, ImpacTheming, Impac
   // -------------------------------------------------------
   ImpacLinking.linkData(DevSettings.buildLinkingConfig(defaults.orgUid, defaults.mnoeUrl));
 
+  // Locale to be used
+  $translate.use('en-AU');
```
alexnoox: Shouldn't you configure a default language instead?

cesar-tonnoir: Everything that's in `/workspace` is just the developer interface, it's like an application that integrates impac-angular, so the result would be the same

cesar-tonnoir:
Thanks guys for the review.
@xaun -> done.
Can you write somewhere in your TODO to add a few lines in DEVELOPER.md about internationalisation? thx

PR #358, opened 2017-06-18
IMPAC-598 threshold improv

xaun:
TODO: 
- [x] Add loading indicator when the KPI is being added

### DIFF
--- a/src/services/highcharts-factory/highcharts-factory.svc.coffee
+++ b/src/services/highcharts-factory/highcharts-factory.svc.coffee
@@ -40,12 +40,12 @@ angular
         @template(@data.series, @options)
         @formatters(@data.labels, @options)
         @todayMarker(@data.labels, @options)
-        @thresholdsMarkers(@options)
       )
       if _.isEmpty(@hc)
         @hc = Highcharts.chart(@id, chartConfig)
       else
         @hc.update(chartConfig)
+      @addThresholds(@data.labels, @options)
       return @
 
     template: (series, options = {})->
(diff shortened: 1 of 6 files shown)


src/services/kpis/kpis.svc.coffee
```diff
@@ -341,7 +341,7 @@ angular
               ImpacEvents.notifyCallbacks(IMPAC_EVENTS.addOrRemoveAlerts)
               angular.extend(kpi, success.data)
               _self.buildKpiWatchables(kpi)
-              _self.show(kpi)
+              _self.show(kpi) if showKpi
```
cesar-tonnoir: :+1: indeed...

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -87,16 +82,32 @@ angular
             y: -5
         }]
 
-    thresholdsMarkers: (options = {})->
-      return {} if _.isEmpty(options.thresholds)
-      lines = []
+    addThresholds: (options = @options)->
```
cesar-tonnoir: Just out of curiosity: did you try to keep the plotLine, adding a zone with color = transparent before today?
I'm just wondering if we can avoid the fake serie and adding all this code?..

cesar-tonnoir: Also, I think that you should pass an argument there to decide whether or not you want to display the threshold only in the future: we may want to display it on the full time period on other widgets

xaun: The plotLine doesn't seem to have a Zones API :\ Ok yes I can add an arg for full-length threshold

xaun:
@cesar-tonnoir please review latest commits - thanks

xaun:
https://github.com/maestrano/maestrano-hub/pull/536

xaun:
Yes, there are two linked PRs on this PR. You need mnoe & mno-hub - they have been merged but make sure you have your dev including the required commits.

For the delete / recreate KPI, you mean the series yeah? well you cannot `chart.update()` for new data, when for example you refresh the widget, its best to remove all, and re-add.

For updating the KPI, I guess I could write a new function to `chart.series[0].update()`, but I just thought as the front-end effect would be the same, redrawing the line, that I would re-use the existing update function, rather than adding logic to the factory to determine whether a kpi was being specifically updated. I could look into this further to see if I can find a better way if you think it's worth it.

PR #359, opened 2017-06-19
Add gulp interval improving CPU perf

xaun:
@cesar-tonnoir improves cpu perf by ~30% running serve. Also removed the bower_components watch as it's pretty unnecessary to be watching all those files just so when a component is bower installed it reloads; can just re-run the task on that case. 

:beers:


xaun:
@alexnoox FYI, incase helpful for mnoe-angular team

PR #367, opened 2017-07-14
[IMPAC-631] use ngStyle when binding values

xaun:
IE does not like angular expressions in html attributes like style. @ouranos


alexnoox:
👍

PR #385, opened 2017-08-14
[IMPAC-648] v1.6.0 - Refactor "data not found" mode

cesar-tonnoir:
- Better styling and wording
- Example mode with stub data
- Update delete widget 'pop-up' to match the new design
- Slight refactor of `WidgetSvc`


cesar-tonnoir:
@xaun this is ready - can you please review?
goes with: https://github.com/maestrano/impac/pull/486

src/components/common/delete-widget/delete-widget.directive.coffee
```diff
@@ -0,0 +1,20 @@
+module = angular.module('impac.components.common.delete-widget',[])
+
+module.directive('commonDeleteWidget', ($templateCache, ImpacWidgetsSvc, ImpacUtilities) ->
```
xaun: For new components, we should be leveraging the angular `.component`. 

```
bindings: {
   # one-way binding
   parentWidget: '<' 
}
...

# lifecycle hooks
ctrl.$ngOnInit = ->
  ctrl.loading = false

```

`.directives` should really only be used now to attach functionality onto an element. For example the `impac-angular/src/components/widgets-common/autofocus/autofocus.directive.coffee` directive.

src/components/common/delete-widget/delete-widget.directive.coffee
```diff
@@ -0,0 +1,20 @@
...
+    scope: {
+      parentWidget: '='
+      onDismiss: '&'
+    }
+    link: (scope, element) ->
+      scope.loading = false
+
+      scope.deleteWidget = ->
+        scope.loading = true
+        ImpacWidgetsSvc.delete(scope.parentWidget)
```
xaun: It should be the parents (`impacWidget`) responsibility to make the backend request to delete the widget. This component should just emit an event via an `onDelete` callback.

src/services/widgets/widgets.svc.coffee
```diff
@@ -115,102 +111,114 @@ angular
...
+      # Init settings
+      unless _.isEmpty(widget.settings)
+        for setting in widget.settings
+          setting.initialize() if angular.isDefined(setting.initialize)
+      # Draw chart
+      widget.format() if angular.isDefined(widget.format)
+
+    # Calls Legacy Impac! or bolt API to render a widget
+    # If no content is returned, the endpoint will be called again with `demo` = `true` to retrieve stub data
+    @show = (widget, refreshCache = false, demo = false) ->
```
xaun: Maybe the demo flag should be within an `opts` arg instead? More flexible with future additions.

karma-min.conf.js
```diff
@@ -33,6 +33,7 @@ module.exports = function(config) {
       'bower_components/angular-xeditable/dist/js/xeditable.js',
       'bower_components/angular-toastr/dist/angular-toastr.tpls.js',
       'bower_components/bootstrap/dist/js/bootstrap.js',
+      'bower_components/ng-tags-input/ng-tags-input.js',
```
xaun: Not sure this needs to be commited?

karma.conf.js
```diff
@@ -33,6 +33,7 @@ module.exports = function(config) {
       'bower_components/angular-xeditable/dist/js/xeditable.js',
       'bower_components/angular-toastr/dist/angular-toastr.tpls.js',
       'bower_components/bootstrap/dist/js/bootstrap.js',
+      'bower_components/ng-tags-input/ng-tags-input.js',
```
xaun: Same here

src/services/widgets/widgets.svc.coffee
```diff
@@ -87,11 +83,11 @@ angular
             # The widget's metadata are updated only if the correct setting is attached to the widget
             if settingKey in wgtSettingsKeys
               angular.extend wgt.metadata, setting.toMetadata()
-              wgt.isLoading = true
-              _self.update(wgt, { metadata: wgt.metadata }).then(
-                (updatedWidget) ->
-                  _self.show(updatedWidget).finally( -> updatedWidget.isLoading = false )
-              )
+              promises.push _self.update(wgt, { metadata: wgt.metadata })
```
xaun: Is the `.then((updatedWidget)-> updatedWidget.isLoading = false ` not needed anymore?

src/components/common/delete-widget/delete-widget.directive.coffee
cesar-tonnoir: :+1: done

src/services/widgets/widgets.svc.coffee
cesar-tonnoir: No, that's an improvement: the loader is now fully managed by the `#show` and `#update` methods in the service.
We don't need to care about it in the controllers or other methods anymore.

karma-min.conf.js
cesar-tonnoir: I think this should have been part of v1.5.6, it keeps appearing every time I run the specs...
As all the other bower_components are committed, wouldn't it be more consistent to add this one as well?

cesar-tonnoir:
@xaun thanks for the review. I've applied your suggestions.
I'm going to merge the PR as this needs to be deployed tonight. Please let me know if you have any other comment.

karma-min.conf.js
xaun: Yeah if it is a bower_component then it should be committed. But as I cannot see it in the bower.json I thought it was a mistake. I just tested though and it's writing it to my karma files on gulp test too.. Must be a sub-dep

src/components/common/delete-widget/delete-widget.directive.coffee
xaun: @cesar-tonnoir I think @alexnoox agrees with me judging by the reaction.. I was going to ask for his input. The `delete-widget` should just be a presentational component. If you apply what you have done to all components, you will find that your codebase makes requests to APIs from all sorts of places. It seems logically for it to encapsulate the request as it's purpose it to delete. But it actually makes much more sense for the widget.component to be the "smart component", and for it to make the request to the widget service. 

Also, think about having many components like this, you will then have the WidgetsSvc dependency across many different places, and it becomes hard to follow.  

You can read up on this design pattern by searching "smart components dumb components" or "smart components presentational components" (e.g http://blog.angular-university.io/angular-2-smart-components-vs-presentation-components-whats-the-difference-when-to-use-each-and-why/).

xaun:
@cesar-tonnoir In my review I forgot to highlight the fact that a config change was made & the README.md configurations section(s) were not updated. Do you think you could patch this? Cheers

xaun:
@cesar-tonnoir also as the `data-not-found.directive` images have been removed, mentions of `defaultImagesPath` also need to be removed. 

This includes:
- readme.md
- assets.svc.coffee
- workspace/app/app.js

PR #393, opened 2017-08-29
[IMPAC-622] threshold tooltip

xaun:
@cesar-tonnoir here is the threshold tooltip rebased on top of v.1.6.1. I have applied your refactors as well. 

I could not assign the switch like you wanted to in the kpi.svc: 
```
host = switch.kpi.source
    ...
```
As one of the cases `return $q.reject(err)`


xaun:
@cesar-tonnoir I have added commits to: 
- [ ] flag impac dependency in Changelog
- [ ] clean up toastr error/success messages

Missing:
- [ ] remove outline from kpi threshold buttons when loading

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -82,22 +87,23 @@ angular
...
         }]
 
-    addThresholds: (options = @options)->
-      return if _.isEmpty(@hc)
+    addThresholds: (chart, options = @options)->
+      return if _.isEmpty(chart)
       # Remove existing thresholds
-      _.each(@hc.series, (s)-> s.remove() if s.name.toLowerCase().includes('threshold'))
-      return @hc if _.isEmpty(options.thresholds)
+      _.each(chart.series, (s)-> s.remove() if s.name.toLowerCase().includes('threshold'))
```
cesar-tonnoir: Do we have to remove all the existing threshold series? What if we want to define several thresholds on a chart in the future?

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -106,8 +112,47 @@ angular
...
         serie.data.push.apply(serie.data, thresholdBar)
         # Note: series can only be added after initial render via the `addSeries` method.
-        @hc.addSeries(serie, true)
-      @hc
+        chart.addSeries(serie, true)
+      chart
+
+    # Render a threshold serie dataLabel when KPI triggered
+    renderThresholdTooltip: (chart, options = @options)->
+      thresholdSerie = _.find(chart.series, (s)->
```
cesar-tonnoir: should be passed as an argument

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -106,8 +112,47 @@ angular
...
+        s.options.triggered && s.name.toLowerCase() == 'threshold kpi'
+      )
+      return removePointDataLabels(@triggeredPoint) unless thresholdSerie
```
cesar-tonnoir: Are you sure we should have `@triggeredPoint` as a global variable? shouldn't it remain in the method's scope?

src/services/kpis/kpis.svc.coffee
```diff
@@ -175,6 +175,12 @@ angular
       templ = _self.getKpiTemplate(kpiEndpoint, kpiWatchable)
       ((templ? && templ.target_placeholders?) && templ.target_placeholders[kpiWatchable]) || {}
 
+    @getApiV2KpiDataKey = (kpi)->
+      # Formats the kpi endpoint to select the key name
+      # e.g response.cash_projection = { triggered: true, ... }
+      # TODO: maybe the 'kpi' of endpoint 'kpis/cash_projection' should be removed?
```
cesar-tonnoir: Yes: I don't think the endpoint was returning `kpis/something` in the API v1?
Can you please align the Bolt endpoint?

PR #397, opened 2017-09-02
[IMPAC-618] Better time management for cashflow widgets

cesar-tonnoir:
- Uses highstock instead of highchart
- Build labels based on timestamps returned with each series instead of dates strings
- Better zooming capabilities
- Change time period using dates pickers
- Better granularity (use daily intervals)

Goes with https://github.com/maestrano/impac-finance-bolt/pull/44

voila_capture 2017-09-03_2-30-45_am

**TODO**
- [x] Fix threshold tooltip
- [ ] Fix threshold panel appears when using the zoom shortcuts

### DIFF
--- a/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
+++ b/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
@@ -70,17 +70,13 @@ module.component('chartThreshold', {
       ctrl.loading = true
       params = targets: {}, metadata: {}
       params.targets[ctrl.kpi.watchables[0]] = [{
-        "#{ctrl.kpiTargetMode}": ctrl.draftTarget.value
+        "#{ctrl.kpiTargetMode}": parseFloat(ctrl.draftTarget.value)
       }]
       return unless ImpacKpisSvc.validateKpiTargets(params.targets)
       promise = if ctrl.isEditingKpi
         ImpacKpisSvc.update(getKpi(), params, false)
       else
-        # TODO: improve the way the hist_params are applied onto widget kpis
-        if ctrl.widget.metadata && (widgetHistParams = ctrl.widget.metadata.hist_parameters)
-          params.metadata.hist_parameters = widgetHistParams
-        else
-          params.metadata.hist_parameters = ImpacUtilities.yearDates()
+        params.metadata.hist_parameters = ctrl.widget.metadata.hist_parameters
         params.widget_id = ctrl.widget.id
         ImpacKpisSvc.create('impac', ctrl.kpi.endpoint, ctrl.kpi.watchables[0], params)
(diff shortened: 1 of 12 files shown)


src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
```diff
@@ -165,7 +161,7 @@ module.component('chartThreshold', {
       widgetHistParams = ctrl.widget.metadata && ctrl.widget.metadata.hist_parameters
       # Widget histParams are YTD by default (when undefined on metadata),
       # therefore in the past by default
-      ctrl.disabled = _.isEmpty(widgetHistParams) || moment(widgetHistParams.to) <= moment().startOf('day')
+      ctrl.disabled = widgetHistParams? && moment(widgetHistParams.to) <= moment.utc().startOf('day')
```
xaun: if `widgetHistParams` is equal to `undefined` or `null`, `ctrl.disabled` will equal `false`. Is this intended?

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -38,7 +42,7 @@ angular
       angular.extend(@options, options)
       chartConfig = angular.merge({}, @template(), @formatters(), @todayMarker())
       if _.isEmpty(@hc)
-        @hc = Highcharts.chart(@id, chartConfig)
+        @hc = Highcharts.stockChart(@id, chartConfig)
```
xaun: Will all Impac! angular charts use StockChart now?

src/components/widgets-common/chart-threshold/chart-threshold.less
```diff
@@ -40,7 +40,14 @@ chart-threshold {
...
     & > .btn.has-spinner {
-      i { padding: 0 15px; }
+      outline: none;
+      i {
+        padding: 0 15px;
+        outline: none;
+      }
+      span {
+        outline: none;
+      }
```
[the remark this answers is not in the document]
cesar-tonnoir: :+1:
cherry-picked your commit

src/services/highcharts-factory/highcharts-factory.svc.coffee
cesar-tonnoir: If I'm not wrong, stockChart is just like highCharts with more options, so I think this should be ok?..

src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
cesar-tonnoir: yes: I disable it only if I have hist parameters defined so that the time period is in the past.
the default time period is no longer YTD but -3m / +1m

PR #398, opened 2017-09-05
[IMPAC-617] Transactions list in Cash Projection widget

cesar-tonnoir:
**Note** check only changes from #fda28a3 as feature/cash-trx-list branches off feature/time-management which is not merged yet

- New BoltResources service to query the Bolt's json api
- New component to display transactions list with pagination
- Displays transactions details when the user clicks on a bar (shows bills when click on "Payables", invoices when click on "Receivables")

voila_capture 2017-09-05_10-35-24_pm

@alexnoox feel free to have a look if you wish ;)

**Dependencies**
- https://github.com/maestrano/impac/pull/497
- https://github.com/maestrano/impac-finance-bolt/pull/45

PR #403, opened 2017-09-06
[IMPAC-641] IE11: Cash Balance & Projection widget

rheasunshine:
@cesar-tonnoir 
-Resizing of cash balance widget fixed

### DIFF
--- a/src/components/widgets/accounts-cash-balance/accounts-cash-balance.less
+++ b/src/components/widgets/accounts-cash-balance/accounts-cash-balance.less
@@ -4,12 +4,12 @@
   .data-container {
     height: ~"calc(@{impac-big-widget-size} - 20px)";
     width: 100%;
-    display: flex;
   }
 
   .left-panel {
-    flex: 0;
-    min-width: 180px;
+    display: inline-block;
+    width: 180px;
+    height: ~"calc(@{impac-big-widget-size} - 20px)";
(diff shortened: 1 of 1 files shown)


src/components/widgets/accounts-cash-balance/accounts-cash-balance.less
```diff
@@ -4,12 +4,12 @@
...
     width: 100%;
-    display: flex;
   }
 
   .left-panel {
-    flex: 0;
-    min-width: 180px;
+    display: inline-block;
+    width: 180px;
+    height: ~"calc(@{impac-big-widget-size} - 20px)";
```
cesar-tonnoir: I came up with the exact same solution in my current branch :)
I don't think you need this line though (`height: ~"calc(@{impac-big-widget-size} - 20px)";`) as the height is already set by `.cash-balance-chart`?

cesar-tonnoir: I came up with the exact same solution in my current branch :)
I don't think you need this line though (height: ~"calc(@{impac-big-widget-size} - 20px)";) as the height is already set by .cash-balance-chart?

cesar-tonnoir:
Actually, no need for you to push more changes on this one => this is going to be fixed by #398

PR #408, opened 2017-09-20
[IMPAC-667] Update time management on cashflow widgets

cesar-tonnoir:
After this change, the user won't be able to modify the time period on the cashflow widgets, but only to change the zooming instead.

Depends on https://github.com/maestrano/impac-finance-bolt/pull/51


xaun:
@cesar-tonnoir should there be a dependency documented in the Changelog?

cesar-tonnoir:
no: the change is backward-compatible

PR #396, opened 2017-09-01
[IMPAC-616] Stacking columns for cash projection overdue invoices

xaun:
screen shot 2017-09-01 at 17 51 45

PR #410, opened 2017-09-21
[IMPAC-616] Stack overdue transactions

cesar-tonnoir:
replaces #396

### DIFF
--- a/src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
+++ b/src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
@@ -9,12 +9,14 @@ module.controller('WidgetAccountsCashProjectionCtrl', ($scope, $q, $filter, Impa
   $scope.datesPickerDeferred = $q.defer()
   $scope.intervalsOffsetsDeferred = $q.defer()
   $scope.currentOffsetsDeferred = $q.defer()
+  $scope.rangesDeferred = $q.defer()
 
   settingsPromises = [
     $scope.orgDeferred.promise,
     $scope.datesPickerDeferred.promise,
     $scope.intervalsOffsetsDeferred.promise,
-    $scope.currentOffsetsDeferred.promise
+    $scope.currentOffsetsDeferred.promise,
+    $scope.rangesDeferred.promise
   ]
 
   # Simulation mode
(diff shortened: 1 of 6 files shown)


src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
```diff
@@ -28,6 +30,14 @@ module.controller('WidgetAccountsCashProjectionCtrl', ($scope, $q, $filter, Impa
     label: 'Get alerted when the cash projection goes below'
   }
 
+  # Define metadata for overdue transactions ranges
+  w.metadata.ranges = [
+    { name: '>60 days', to: moment().subtract(61,'d').format('YYYY-MM-DD') },
+    { name: '30-60 days', from: moment().subtract(60,'d').format('YYYY-MM-DD'), to: moment().subtract(30,'d').format('YYYY-MM-DD') },
+    { name: '<30 days', from: moment().subtract(29,'d').format('YYYY-MM-DD'), to: moment().format('YYYY-MM-DD') }
+  ]
```
xaun: Do these params get applied when you save the widget settings? That was the reason I added the `toMetadata` & `initialize` method.

src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
```diff
@@ -28,6 +30,14 @@ module.controller('WidgetAccountsCashProjectionCtrl', ($scope, $q, $filter, Impa
...
+  $scope.rangesDeferred.resolve()
```
xaun: I cant see anything using the deferred? Why do we have it?

src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
```diff
@@ -130,7 +141,7 @@ module.controller('WidgetAccountsCashProjectionCtrl', ($scope, $q, $filter, Impa
       currency: w.metadata.currency
       showToday: true
       showLegend: true
-      thresholds: getThresholds()
+      stacking: 'normal'
```
xaun: This should be removed as it is send via the API

src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
xaun: Looks like they are being applied now =\

src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
xaun: Removing this deferred seems to work, am I missing something?

PR #420, opened 2017-10-05
[IMPAC-663] Fix unwanted triggering of create threshold panel

xaun:
Some things to note:
- `Highcharts.addEvent(chart.hc.container, 'click', onChartClick)` applies the click event to the whole canvas, causing the need for the existing (incomplete) guarding in the `onChartClick` callback method. 
- The native `chart.events.click` callback only triggers on chart clicks. Fixing the problem of needing to guard for things like legend, zoom bar, etc clicks.
- The native `chart.events.click` callback can not be updated after render.

### DIFF
--- a/src/services/highcharts-factory/highcharts-factory.svc.coffee
+++ b/src/services/highcharts-factory/highcharts-factory.svc.coffee
@@ -9,6 +9,8 @@ angular
           type: 'line'
           zoomType: 'x'
           spacingTop: 20
+          events:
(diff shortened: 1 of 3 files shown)


src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -9,6 +9,8 @@ angular
           type: 'line'
           zoomType: 'x'
           spacingTop: 20
+          events:
```
cesar-tonnoir: :+1:

PR #422, opened 2017-10-12
[IMPAC-614] Add editable expected payment date to transactions list

cesar-tonnoir:
- Filter transactions on expected payment date instead of due date
- Update expected payment date for each transaction
- Fix timezone-related errors


src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -1,7 +1,7 @@
 angular
 .module('impac.services.highcharts-factory', [])
 .factory('HighchartsFactory', ($filter)->
-
+
```
alexnoox: Remove empty spaces

cesar-tonnoir: :+1: will do with the next change

PR #424, opened 2017-10-17
Dynamic date format

Vladimir31415:
Added Date-formatter service.
In theming service there is a section for date-formatting. Usually widgets use dates-picket or time-period component so there are formats for these components. But some widgets use dates outside of these components so it's possible to specify format for single component.
In common case date-formatter service will use locale based on  angular-translate locale


src/components/widgets-settings/dates-picker/dates-picker.directive.coffee
```diff
@@ -85,23 +90,26 @@ module.directive('settingDatesPicker', ($templateCache, $filter, ImpacWidgetsSvc
...
+#
+#          if Date.parse(scope.toDate) && !scope.keepToday
+#            parsedTo = scope.toDate.split('-')
+#            y = parsedTo[0]
+#            m = parsedTo[1] - 1
+#            d = parsedTo[2]
+#            scope.calendarTo.value = new Date(y,m,d)
+#          else
+#            scope.calendarTo.value = new Date()
+          scope.calendarFrom.value = moment(scope.fromDate, scope.dateFormat).toDate()
```
cesar-tonnoir: :+1: 
Can you remove the comments if obsolete?

src/services/theming/theming.svc.coffee
```diff
@@ -126,6 +126,18 @@ angular
         tagging:
           enabled: false
 
+      # date-formatter settings for widgets
+      dateFormatterSettings:
```
cesar-tonnoir: I like the idea there, but I think that by default, all the widgets should use the same format. We could have something like:
```coffeescript
# Default specified in ThemingSvc
dateFormatterSettings:
  default: 'L'
  widgets: {}
  settings: {}

# Custom configuration defined in the host app
dateFormatterSettings:
  default: 'YYYY-MM-DD'
  widgets:
    'hr/employee_details': 'MM-DD-YYYY'
  settings:
    'dates-picker': 'DD/MM/YYYY'
```

Behaviour:
- if `dateFormatterSettings.widgets['hr/employee_details']` if defined => use the defined format
- otherwise, use `dateFormatterSettings.default`

src/services/date-formatter/date-formatter.svc.coffee
```diff
@@ -0,0 +1,41 @@
...
+  .module('impac.services.date-formatter', [])
+  .service 'ImpacDateFormatter', ($log, $locale, ImpacTheming, $translate) ->
+
+    @getFormatForEntity = (entity = '') ->
+
+      settings = ImpacTheming.get()
+      if settings.dateFormatterSettings && settings.dateFormatterSettings[entity]
+        format = settings.dateFormatterSettings[entity]
+      else
+        format = 'L'
```
cesar-tonnoir: see comment in `ThemingSvc` => default should be fetched from the config

src/services/date-formatter/date-formatter.svc.coffee
```diff
@@ -0,0 +1,41 @@
...
+      return format
+
+    @formatDateString = (obj, widgetEndpoint = '') ->
```
cesar-tonnoir: This method is very obscure... :/ I think you are trying to perform a generic conversion of all the date members of any object (tell me if I'm wrong).
However I don't think this is the way to go: what if your object is something like:
```coffeescript
object:
  content:
    values:
      dates: ['2015-05-01', '2015-06-01']
```
? I don't think the dates would be converted in this case.

Instead, I would suggest to format the date **when it is displayed**, directly in the view.

src/services/dashboards/dashboards.svc.coffee
```diff
@@ -110,12 +112,20 @@ angular
...
-                  _self.setDashboards(response.data).then(-> _self.setCurrentDashboard() )
-                (error) ->
-                  $log.error("Impac! - DashboardsSvc: cannot retrieve dashboards list for org: #{orgId}")
-              )
+              dashboardsPromise = $http.get(ImpacRoutes.dashboards.index(orgId))
+                .then(
+                  (success) ->
+                    success.data.forEach (o) ->
+                      o.widgets.forEach (widget) ->
+                        ImpacDateFormatter.formatDateString widget, widget.endpoint
```
cesar-tonnoir: This is too intrusive. See comment in `DatesFormatterSvc` => I think the formatting should be done in the views rather than directly on the object.

src/services/dashboards/dashboards.svc.coffee
```diff
@@ -157,6 +167,8 @@ angular
       widgetsTemplatesPromises.push $http.get(widgetsTemplatesUrl()).then(
         (response) ->
           for widgetTemplate in response.data.widgets
+            if _.get(widgetTemplate, 'metadata.hist_parameters.from') && _.get(widgetTemplate, 'metadata.hist_parameters.to')
```
cesar-tonnoir: Same here. Plus, remember the `metadata` objects will be sent to the Impac! API. You can't assume the API will accept any date format...

src/services/utilities/utilities.svc.coffee
```diff
@@ -66,6 +66,7 @@ angular
 
         if histParams.from
           fromDate = histParams.from
+#          fromDate = moment(histParams.from).format("DD-MM-YYYY")
```
cesar-tonnoir: Why don't you use your formatting service in this file?

src/services/widgets/widgets.svc.coffee
```diff
@@ -189,30 +190,36 @@ angular
...
-                # Reload the widget with param `demo` set at `true` (to retrieve stub data)
-                if demoData
-                  # If we were already trying to load the widget in demo mode, we resolve the widget without content
-                  # TODO: display an error box?
-                  $log.error('Impac! - WidgetsSvc: Cannot retrieve demo data for widget:', widget)
-                  $q.resolve(widget)
+          $http.get(url, config)
+            .then(
+              (success) ->
+                ImpacDateFormatter.formatDateString success.data, widget.endpoint
```
cesar-tonnoir: Same comment as per dashboards svc: this is too intrusive for the existing code. Please apply the formatting in the view directly

Vladimir31415:
Thanks for your review.
As i understood your proposal is to make changes (use momentFormat filter for displaying) in all current widget views. And every coming widget will use this filter in its view. Right?

cesar-tonnoir:
@Vladimir31415 yes, that's the idea.

Vladimir31415:
The update is added.  
In order not to add extra parameter for filter, dateFormatterSettings object in theming,svc has structure
```
dateFormatterSettings: {
    default: 'L',
    formats: {
        # widgets
        # common components
        # filters
    }
}
```
Too intrusive 'interceptor' is removed.
MomentDate filter is used for formatting dates.

cesar-tonnoir:
Hi @Vladimir31415 ,
It looks like most of my previous comments are still valid unfortunately:
- You can't modify the date format directly on the dashboard/widgets objects as you currently do in `DashboardsSvc.fetchWidgetsTemplates`. Actually this feature should not require any change on DashboardsSvc / WidgetsSvc / KpisSvc, etc.
- The `formatDateString` method is far too complex, we need a different approach: the date formatter should be a simple filter that takes a String as an argument and formats it depending on the configuration available in `ThemingSvc`. The dates should be formatted **in the views** as stated before: this means that every html template where a date is displayed should be using your filter.
- Having different formats in the `dateFormatterSettings` is a good idea, but the "special" formats should all be commented: we want all the dates to look the same by default
- Several filters are currently used to format dates in the project (`mnoDate`, `momentDate`, `date`) + one service (`DatesFormatterSvc`) => Only one filter should be used everywhere if we want the dates to be consistent, and I'm not convinced you need one filter AND one service (what prevents you to put all the code in the filter directly?)

cesar-tonnoir:
Hi @Vladimir31415 
Thank you for this latest change - it looks good to me, the different points mentioned previously seem to be addressed so we are going to send this PR to QA testing next week.
Cheers! :beer:

PR #426, opened 2017-10-19
Fix empty sso being applied to widget show headers

xaun:


xaun:
@cesar-tonnoir should this be merged into the next 1.6 release?

PR #456, opened 2018-01-22
[IMPAC-758] Date hidden in X-axis for current ratio and debt service widget

MAhsenArif:


src/components/widgets/accounts-ratios/accounts-ratios.directive.coffee
```diff
@@ -33,6 +33,7 @@ module.controller('WidgetAccountsRatiosCtrl', ($scope, $q, $filter, $translate,
   $scope.intervalsCount = 0
   $scope.isPnl = false
   $scope.periodInfoContext = {}
+  $scope.allowXAxisLegend = true
```
xaun: No need to initialise the property here as well as in the `w.initContext` initializer, please remove.

src/components/widgets/accounts-ratios/accounts-ratios.directive.coffee
```diff
@@ -72,6 +74,9 @@ module.controller('WidgetAccountsRatiosCtrl', ($scope, $q, $filter, $translate,
         $scope.calculatedNumerator = _.last(w.content.calculation.numerator.values)
         $scope.calculatedDenominator = _.last(w.content.calculation.denominator.values)
 
+  $scope.allowWidgetLegend = ->
```
xaun: This method is not being used in the view, so it should not be bound to $scope.

src/components/widgets/accounts-ratios/accounts-ratios.directive.coffee
```diff
@@ -57,6 +58,7 @@ module.controller('WidgetAccountsRatiosCtrl', ($scope, $q, $filter, $translate,
       $scope.periodInfoContext.histParams = w.metadata.hist_parameters
       $scope.periodInfoContext.accountingBehaviour = behaviour
       getPrefix(behaviour).then((label)-> $scope.periodInfoContext.injectBefore = label)
+      $scope.allowXAxisLegend = $scope.allowWidgetLegend()
```
xaun: As this method is only being invoked here, I think we can just do this:
```
$scope.allowXAxisLegend = !_.includes(['accounts/ratios/current', 'accounts/ratios/debt_service'], w.endpoint)
```
This leverages Lodash, to shorten & add readability to the statement, and then we can remove the `allowWidgetLegend` method.

src/components/widgets/accounts-ratios/accounts-ratios.tmpl.html
```diff
@@ -86,7 +86,7 @@ <h4>Widget settings</h4>
 
         <div class="history chart-container" ng-show="widget.isHistoryMode">
           <div impac-chart draw-trigger="::drawTrigger.promise" deferred="::chartDeferred"></div>
-          <div class="legend">{{ widget.content.layout.ratio }}</div>
+          <div ng-show="allowXAxisLegend" class="legend">{{ widget.content.layout.ratio }}</div>
```
xaun: Prefer `ng-if` over `ng-show`. `ng-show` is really only useful if you want to specifically want to keep an element in the DOM.

xaun:
Merged in https://github.com/maestrano/impac-angular/pull/472 to fix conflicts

PR #457, opened 2018-01-23
[IMPAC-757] Added radio mode for organization

MAhsenArif:
Radio mode can be set on Widgets by adding parameter `organization-mode="'single'" ` to `setting-organizations` directive.

### DIFF
--- a/src/components/widgets-settings/organizations/organizations.directive.coffee
+++ b/src/components/widgets-settings/organizations/organizations.directive.coffee
@@ -1,5 +1,5 @@
 module = angular.module('impac.components.widgets-settings.organizations',[])
-module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc) ->
+module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc, ImpacMainSvc) ->
 
   w = $scope.parentWidget
   w.selectedOrganizations = {}
@@ -19,7 +19,7 @@ module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc)
     if multiOrgMode()
       w.selectedOrganizations[orgUid] = !w.selectedOrganizations[orgUid]
       $scope.onSelect({orgs: w.selectedOrganizations}) if angular.isDefined( $scope.onSelect )
-    if singleOrgMode()
+    else if singleOrgMode()
       angular.forEach w.selectedOrganizations, (value, key) ->
         w.selectedOrganizations[key] = false
       w.selectedOrganizations[orgUid] = true
@@ -34,16 +34,23 @@ module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc)
     ImpacDashboardsSvc.load().then(
       (config) ->
         $scope.dashboardOrganizations = config.currentDashboard.data_sources
+
         if w.metadata? && w.metadata.organization_ids?
-          # Note: For a widget in a dashboard multiple companies, we select the
-          #       first company by default ONLY if the mode is radio.
           count = 0
+          widgetOrgIds = w.metadata.organization_ids
+          currentOrganization = ImpacMainSvc.config.currentOrganization
+
           for org in $scope.dashboardOrganizations
-            orgSelection = _.contains(w.metadata.organization_ids, org.uid)
+            orgSelection = _.contains(widgetOrgIds, org.uid)
             w.selectedOrganizations[org.uid] = orgSelection
-            if singleOrgMode() && orgSelection
-              w.selectedOrganizations[org.uid] = if count >= 1 then false else true
-              count += 1
+            if singleOrgMode()
+              if widgetOrgIds.length > 1
+                w.selectedOrganizations[org.uid] = false
+              else if orgSelection
+                w.selectedOrganizations[org.uid] = if count >= 1 then false else true
+                count += 1
+          if singleOrgMode() && widgetOrgIds.length > 1
+            w.selectedOrganizations[currentOrganization.uid] = true
           setting.isInitialized = true
     )
 
--- a/src/components/widgets-settings/organizations/organizations.directive.coffee
+++ b/src/components/widgets-settings/organizations/organizations.directive.coffee
@@ -35,20 +35,19 @@ module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc,
       (config) ->
         $scope.dashboardOrganizations = config.currentDashboard.data_sources
 
-        if w.metadata? && w.metadata.organization_ids?
-          widgetOrgIds = w.metadata.organization_ids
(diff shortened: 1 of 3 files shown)


src/components/widgets-settings/organizations/organizations.directive.coffee
```diff
@@ -21,9 +34,23 @@ module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc)
     ImpacDashboardsSvc.load().then(
       (config) ->
         $scope.dashboardOrganizations = config.currentDashboard.data_sources
+
         if w.metadata? && w.metadata.organization_ids?
+          count = 0
+          widgetOrgIds = w.metadata.organization_ids
+          currentOrganization = ImpacMainSvc.config.currentOrganization
```
xaun: This block could do with some refactoring, simplifying and commenting. The code is not very readable, and will cause future confusion for developers.

After looking into for a bit, Im not even sure it's necessary to be iterating over all the `$scope.dashboardOrganizations` and cramming all the logic into all the `if`'s & `else`'s, which Im not even sure we need all of.

I have tried to capture all the cases you have coded for in comments:
```coffeescript
# single mode 
# - has multiple data sources previously selected 
# => deselect all & select current organization
# - none or 1 data source previously selected
# => apply the previously first widget data source organization, or select the first one
```
Is this correct? 

We shouldn't have to "deselect" (e.g set all the key values to false), as this is inside an initialize method, no "checkboxes" are going to be ticked, right?  This saves us having to iterate and set everything to false, and having to use a counter.

The thing with an `ng-model` binding, is that if it is `undefined` or `null`, it's falsey. We should only need to set it when it's been toggled for the first time.

I may not have capture all the cases, but this is a quick proposition of a refactor:

```coffeescript
if w.metadata? && w.metadata.organization_ids?
  if singleOrgMode()
    currentOrganization = ImpacMainSvc.config.currentOrganization
    unless _.isEmpty(currentOrganization)
      w.selectedOrganizations[currentOrganization.uid] = true
    else if (dhbOrg = $scope.dashboardOrganizations[0])?
      w.selectedOrganizations[dhbOrg.uid] = true
```

Anyway I may be completely missing something, but if you could just define & code the logic in a clearer manner that would be awesome! Thanks :)

src/components/widgets-settings/organizations/organizations.directive.coffee
```diff
@@ -1,15 +1,28 @@
 module = angular.module('impac.components.widgets-settings.organizations',[])
-module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc) ->
+module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc, ImpacMainSvc) ->
 
   w = $scope.parentWidget
   w.selectedOrganizations = {}
 
+
+  $scope.organizationMode ||= 'multiple'
```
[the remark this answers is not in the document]
MAhsenArif: @xaun  Currently I've have added the ability to use `single` mode but yes it's not being used anywhere. As per discussion with @cesar-tonnoir , this is something the developers will activate when coding a specific widget.

I've also added a message in PR description on how it can be used for a widget.

src/components/widgets-settings/organizations/organizations.directive.coffee
```diff
@@ -1,15 +1,28 @@
...
+  mode = $scope.organizationMode
```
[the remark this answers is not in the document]
MAhsenArif: Added it as a bit of readability. `$scope.organizationMode` was too long to be used everywhere :)

src/components/widgets-settings/organizations/organizations.directive.coffee
```diff
@@ -36,21 +36,18 @@ module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc,
         $scope.dashboardOrganizations = config.currentDashboard.data_sources
 
         if w.metadata? && w.metadata.organization_ids?
-          count = 0
           widgetOrgIds = w.metadata.organization_ids
-          currentOrganization = ImpacMainSvc.config.currentOrganization
+          if singleOrgMode()
+            resetSelectedOrgs()
+            if _.map(widgetOrgIds, -> true).length > 1
```
xaun: why not just `if widgetOrgIds.length > 1`? What's the `_.map` for? If `w.metadata.organization_ids` is defined, it should always be an Array.

src/components/widgets-settings/organizations/organizations.directive.coffee
```diff
@@ -36,21 +36,18 @@ module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc,
         $scope.dashboardOrganizations = config.currentDashboard.data_sources
 
         if w.metadata? && w.metadata.organization_ids?
```
xaun: Might be nice to flatten this a bit by going:
```coffeescript
return unless w.metadata? && w.metadata.organization_ids?
```

Also I've noticed at the bottom of this callback, `setting.isInitialized = true` is nested within the `if` block, so if there are no `w.metadata.organization_ids`, it will never set itself as `isInitialized`. 

Not sure if this is intended or not. But I think maybe we could move this into a `.finally()` callback?

```coffeescript
ImpacDashboardSvc.load().then(
    (config)->
         ...
).finally(->
   settings.isInitialized = true
)
```

What do you think?

src/components/widgets-settings/organizations/organizations.directive.coffee
```diff
@@ -21,9 +34,20 @@ module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc)
...
+          widgetOrgIds = w.metadata.organization_ids
+          if singleOrgMode()
+            resetSelectedOrgs()
+            if _.map(widgetOrgIds, -> true).length > 1
+              currentOrganization = ImpacMainSvc.config.currentOrganization
+              w.selectedOrganizations[currentOrganization.uid] = true
+            else
+              w.selectedOrganizations[widgetOrgIds[0]] = true
+
+          if multiOrgMode()
```
xaun: Maybe use `else if` like you have in the `$scope.toggleSelectOrganization` method.

src/components/widgets-settings/organizations/organizations.directive.coffee
MAhsenArif: @xaun Agree with setting `isInitialized` in `finally` block. Though I'm not yet sure what it's purpose is. I searched for it but it is not used in any views.

xaun: @MAhsenArif good point, it seems only the widgets-settings/accounts-list setting uses it in its `settings.toMetadata` method. 

I think this was an old framework specific convention of sorts.. @cesar-tonnoir can you confirm if there is any reason to keep this property? It looks like our newer widgets-settings components are not implementing it? If not, we should remove these from the old widgets.

PR #458, opened 2018-01-24
[MARLIN-8] Add button among top bar to export as CSV

manu-d:


src/components/widgets-common/top-buttons/top-buttons.directive.coffee
```diff
@@ -16,6 +16,20 @@ module.controller('CommonTopButtonsCtrl', ($scope, $rootScope, $log, ImpacWidget
 
   $scope.hasInfo = ->
     w && w.content? && w.content.info? && w.content.info.length > 0
+
+  $scope.isCsvExportable = ->
```
cesar-tonnoir: After your PR on the bolt is updated, you should just check if your widget is a bolt widget: all the bolt widget should be "csv_exportable"

src/components/widgets-common/top-buttons/top-buttons.directive.coffee
```diff
@@ -16,6 +16,20 @@ module.controller('CommonTopButtonsCtrl', ($scope, $rootScope, $log, ImpacWidget
...
+    templates = _.filter(ImpacDashboardsSvc.getWidgetsTemplates(), {name: w.name})
```
xaun: `w.name` is an editable field, I suggest using `w.endpoint` to find the correct template, as it's the most consistent identifier for templates. 

Also, rather than using `_.filter`, then `templates[0]`, can you just use `template =  _.find(..)` ?

src/components/widgets-common/top-buttons/top-buttons.directive.coffee
```diff
@@ -16,6 +16,20 @@ module.controller('CommonTopButtonsCtrl', ($scope, $rootScope, $log, ImpacWidget
...
+    templates.length > 0 && templates[0].isBolt
```
xaun: Rather than adding the `isBolt` property to the template objects. I think you could just go:
```
templates.length > 0 && _.get(template, 'metadata.bolt_path')
```

cesar-tonnoir:
@manu-d I think your changes on WidgetsSvc are breaking the build => can you please check TravisCi and fix?

PR #459, opened 2018-01-24
[IMPAC-751] Fixed Multi-Company Ordering

iseessel:
@cesar-tonnoir 

I was able to order the settings internally in the directives -- however, the widget-specific data was not uniform enough to do this at the directive level. Since lots of widgets reuse the same html file, I did not have to add that many orderBys.

### DIFF
--- a/src/components/widgets-settings/organizations/organizations.directive.coffee
+++ b/src/components/widgets-settings/organizations/organizations.directive.coffee
@@ -32,7 +32,8 @@ module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc,
   setting.initialize = ->
     ImpacDashboardsSvc.load().then(
       (config) ->
-        $scope.dashboardOrganizations = config.currentDashboard.data_sources
+        currentDashboard = config.currentDashboard
+        $scope.dashboardOrganizations = _.sortBy(currentDashboard.data_sources, 'label')
 
         return unless w.metadata? && w.metadata.organization_ids?
         widgetOrgIds = w.metadata.organization_ids
--- a/src/components/widgets/accounts-balance-sheet/accounts-balance-sheet.directive.coffee
+++ b/src/components/widgets/accounts-balance-sheet/accounts-balance-sheet.directive.coffee
@@ -71,16 +71,16 @@ module.controller('WidgetAccountsBalanceSheetCtrl', ($scope, $q, ImpacWidgetsSvc
   w.initContext = ->
     # if w.content? && w.content.period? && _.contains(_.pluck($scope.periodOptions, 'value'), w.content.period)
     #   $scope.period = angular.copy _.find $scope.periodOptions, ((o) -> o.value == w.content.period)
-
     if $scope.isDataFound = angular.isDefined(w.content) && !_.isEmpty(w.content.summary) && !_.isEmpty(w.content.dates)
       $scope.dates = w.content.dates
       $scope.unCollapsed = w.metadata.unCollapsed || []
 
-      $scope.categories = []
-      translateCategories(Object.keys(w.content.summary))
-
+      translateCategories(Object.keys(w.content.summary)).then((categories) ->
+        # Must sort data after categories have been populated.
+        $scope.categories = categories
+        sortData()
+      )
     initDates()
-    sortData()
 
   $scope.toggleCollapsed = (categoryName) ->
     if categoryName?
@@ -100,20 +100,20 @@ module.controller('WidgetAccountsBalanceSheetCtrl', ($scope, $q, ImpacWidgetsSvc
         return true
 
   translateCategories = (categories) ->
-    _.each(categories, (category) ->
+    translationPromises = _.map(categories, (category) ->
       $translate('impac.widget.account_balance_sheets.' + category.toLowerCase()).then(
         (translation) ->
-          $scope.categories.push({label: translation, key: category})
+          {label: translation, key: category}
 
-        (translationId) ->  # If there is no translation, keep the original
-          $scope.categories.push({label: category.toLowerCase(), key: category})
+        () -> # If there is no translation, keep the original
+          {label: category.toLowerCase(), key: category}
+        )
       )
-      return
-    )
+    $q.all(translationPromises)
 
   sortAccountsBy = (getElem) ->
     angular.forEach($scope.categories, (cat) ->
-      sElem = w.content.summary[cat]
+      sElem = w.content.summary[cat.key]
       if sElem.accounts
         sElem.accounts.sort (a, b) ->
           res = if getElem(a) > getElem(b) then 1
@@ -124,6 +124,7 @@ module.controller('WidgetAccountsBalanceSheetCtrl', ($scope, $q, ImpacWidgetsSvc
     )
 
   sortData = ->
+    return if _.isEmpty($scope.categories)
     if $scope.sortedColumn == 'account'
       sortAccountsBy( (el) -> el.name )
     else if $scope.sortedColumn == 'total1'
--- a/src/components/widgets/accounts-assets-liability-summary/accounts-assets-liability-summary.tmpl.html
+++ b/src/components/widgets/accounts-assets-liability-summary/accounts-assets-liability-summary.tmpl.html
@@ -19,7 +19,7 @@
       <div class="legend">
         <div class="title" ng-show="widget.metadata.organization_ids.length==1">{{widget.content.summary[0].company}} {{classification}}</div>
         <div class="title" ng-hide="widget.metadata.organization_ids.length==1">{{classification}} repartition</div>
-        <span ng-repeat="valuePair in dataSource">
+        <span ng-repeat="valuePair in dataSource | orderBy:'label'">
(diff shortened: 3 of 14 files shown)


src/components/widgets/accounts-balance-sheet/accounts-balance-sheet.directive.coffee
```diff
@@ -100,20 +99,27 @@ module.controller('WidgetAccountsBalanceSheetCtrl', ($scope, $q, ImpacWidgetsSvc
         return true
 
   translateCategories = (categories) ->
+
     _.each(categories, (category) ->
+      # $scope.categories needs to be filled with keys before sortData() runs.
+        # Pushing the categories after $translate resolves would be too late.
+      newCategory = {key: category}
+      $scope.categories.push(newCategory)
```
xaun: I think we can avoid having future developers proceed with caution around this, by maybe having a look at how the component is initialised, and handling the asynchronicity of this method with promises.

When you say "$scope.categories needs to be filled with keys before sortData() runs", is that on load? If so, do you think it would be worth doing something like below:

```coffeescript
  w.initContext = ->
    if $scope.isDataFound = angular.isDefined(w.content) && !_.isEmpty(w.content.summary) && !_.isEmpty(w.content.dates)
      ...
      # While we are here I think we should remove this for the pattern below instead
      # $scope.categories = []
      # translateCategories should use _.map not _.each to return an Array of translated categories
      translateCategories(Object.keys(w.content.summary)).then((categories)->
          $scope.categories = categories
      ).finally(->
        initDates()
        sortData()
      )
    else
      initDates()
      sortData() 
```

I can see `$scope.sort` calls the sortData function, so might have to add some guarding for that as to only run once w.initContext has finished like so:

```coffeescript
sortData = -> 
  return if _.isEmpty($scope.categories)
  ...
```

src/components/widgets-settings/organizations/organizations.directive.coffee
```diff
@@ -16,11 +16,27 @@ module.controller('SettingOrganizationsCtrl', ($scope, $log, ImpacDashboardsSvc)
...
+    )
+
   # initialization of selected organizations
   setting.initialize = ->
     ImpacDashboardsSvc.load().then(
       (config) ->
-        $scope.dashboardOrganizations = config.currentDashboard.data_sources
+        # $scope.dashboardOrganizations is used for the source, as well as each widget's settings.
+        currentDashboard = config.currentDashboard
+        sortedOrgs = sortOrganizations(currentDashboard.data_sources)
```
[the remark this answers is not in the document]
iseessel: Yup agreed!

src/components/widgets/accounts-balance-sheet/accounts-balance-sheet.directive.coffee
iseessel: I'll have a look -- but do agree that handling this with promises is more semantic.

src/components/widgets/accounts-balance-sheet/accounts-balance-sheet.directive.coffee
```diff
@@ -100,20 +99,27 @@ module.controller('WidgetAccountsBalanceSheetCtrl', ($scope, $q, ImpacWidgetsSvc
...
       $translate('impac.widget.account_balance_sheets.' + category.toLowerCase()).then(
         (translation) ->
-          $scope.categories.push({label: translation, key: category})
+          #newCategory is scoped in even though this is a callback.
+          newCategory.label = translation
 
         (translationId) ->  # If there is no translation, keep the original
-          $scope.categories.push({label: category.toLowerCase(), key: category})
+          newCategory.label = translation
```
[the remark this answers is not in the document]
iseessel: Oops, yes it should be `newCategory.label = category.toLowerCase()`. 

But I agree that this fix is not the most semantic and should

src/components/widgets-settings/organizations/organizations.directive.coffee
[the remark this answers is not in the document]
iseessel: Yup!

iseessel:
@xaun Let me know what you think. 

In translateCategories, I made an array of all translate promises, and then called `Promises.all(translationPromises)`. So that only after each promise is fulfilled, sortData is called.

src/components/widgets/accounts-balance-sheet/accounts-balance-sheet.directive.coffee
```diff
@@ -71,16 +71,16 @@ module.controller('WidgetAccountsBalanceSheetCtrl', ($scope, $q, ImpacWidgetsSvc
...
-      $scope.categories = []
-      translateCategories(Object.keys(w.content.summary))
-
-    initDates()
-    sortData()
+      translateCategories(Object.keys(w.content.summary)).then((categories) ->
+        # Must sort data after categories have been populated.
+        $scope.categories = categories
+        initDates()
+        sortData()
```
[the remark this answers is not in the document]
iseessel: Will look into this. Might be best to call initDates outside of the if block like before  (it does not need to wait for the translations), but keep sortDate here.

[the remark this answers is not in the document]
xaun: Yeah that sounds right, also, you shouldn't need to move sortData into a finally. As the rejection has been handled in the translation error handler. When chaining promises, a return in an error handler chains to a successhandler for further .then's down the chain... for e.g:
```coffeescript
asyncFunc = ->
   $http.get('/stuff').then(
      (res)-> 
        return 'yay';
      (err) ->
         # as $http.get has *rejected* the promise, it must be caught. By providing this callback, we have "handled the rejection", and the return value from here will be received as a success to later chained .thens.
       return 'foobar'
   )

asyncFunc().then((res)-> 
      console.log(res)
      # if $http.get errors res = 'foobar', if success, res = 'yay'
)
```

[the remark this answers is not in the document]
iseessel: Yup, that's what I thought, but wasn't entirely sure.

iseessel:
@xaun I moved initDates out from the if block. `#initDates` depends on `w.metadata` which is populated regardless of whether or not impac's requests were successful.

xaun:
@iseessel looks good, can you rebase and fixup your "formatting" and "clean up" commits. It's best to only push commits to remote that provide useful chunks of history. For example when Im working I may make quick commits to save my progress, and try things out, but for a PR, I use rebase to clean up my commit history. 

Not sure if you know, but during rebase, you can even re-order your commits, saving them in that order, or then squashing them up into previous commits. If you use vim as your rebase editor, to move a line it is `:m2` (in visual mode), the number being the index of the line. This is useful as maybe I realised I need to make a cleanup for a commit I made previously, its more relevant to squash that cleanup into a commit 4 commits ago.

iseessel:
@xaun 
Thanks for the help! Let me know how this looks. Tried to keep it across several commits as there were a lot of changes to different parts.

src/components/widgets/accounts-balance-sheet/accounts-balance-sheet.directive.coffee
```diff
@@ -100,20 +100,20 @@ module.controller('WidgetAccountsBalanceSheetCtrl', ($scope, $q, ImpacWidgetsSvc
...
-    _.each(categories, (category) ->
+    translationPromises = _.map(categories, (category) ->
       $translate('impac.widget.account_balance_sheets.' + category.toLowerCase()).then(
         (translation) ->
-          $scope.categories.push({label: translation, key: category})
+          {label: translation, key: category}
 
-        (translationId) ->  # If there is no translation, keep the original
-          $scope.categories.push({label: category.toLowerCase(), key: category})
+        (translationId) ->
```
xaun: remove unused variable `translationId`, add comment back to above line below

src/components/widgets/accounts-balance-sheet/accounts-balance-sheet.directive.coffee
```diff
@@ -100,20 +100,20 @@ module.controller('WidgetAccountsBalanceSheetCtrl', ($scope, $q, ImpacWidgetsSvc
...
+          {label: category.toLowerCase(), key: category}
+        )
       )
-      return
-    )
+    Promise.all(translationPromises)
```
xaun: Probs best to continue using `$q` for promises instead of the JS `Promise`. As it would be the first use in impac-angular. 
```
$q.all(translationPromises)
```

The behaviour will be the same.

xaun: Also I think we are missing a closing parenthesis for the `_.map(` statement

iseessel: Not sure there is? First parenthesis is for the .then(, and second for the _.map(

xaun: Ah ok, sorry it looked weird in my diff

xaun:
@iseessel you'll need to merge release/1.6.9 into your branch or rebase on-top of release/1.6.9, and resolve the conflicts

iseessel:
@xaun  We are merging these changes onto 1.6.9 now?

PR #461, opened 2018-01-26
[IMPAC-745] Fixed Cash Projection Crash

iseessel:
@cesar-tonnoir 
So basically what I did was instead of calling update on highCharts multiple times; each time something is changed in the chart, a new chart is made. For some reason with the highCharts API the create method is faster than the update. 

Because of this, I had to change the way the options were stored, as there are more options at the get go that we need to give to the creation of the graphs.

There was also an issue that if the input was invalid, the buttons wouldn't go back to normal. That was solved by adding a tostr error and invoking cancelCreateKpi. I didn't know what input message would be best so I just put  Please choose a number one or greater(as this would fail when the number was 0).

### DIFF
--- a/src/components/chart/chart.directive.coffee
+++ b/src/components/chart/chart.directive.coffee
@@ -46,7 +46,7 @@ angular
             $log.error(error)
           (chartData) ->
             userAgent = $window.navigator.userAgent
-            
+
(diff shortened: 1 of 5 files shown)


iseessel:
See most recent pull request.

PR #463, opened 2018-01-26
[IMPAC-745] Fixed Cash Projection Crash 

iseessel:
@cesar-tonnoir

Instead of calling update on highCharts multiple times new data is introduced, I created an entirely new chart. For some reason I could not figure out the create method is faster than the update for the highCharts API.

Because of this, I had to change the way the options were stored, as we need to put all the options in at once.

Additionally if an input was invalid, the create and cancel buttons would not go back to normal. I solved that by adding a tostr error and invoking #cancelCreateKpi. I don't know what the intended behavior was, but you could not choose a number less than 0.

I am also not entirely confident in how the options from the specific widgets is getting integrated into the highCharts factory. It seems like _.get is being used to retrieve options from the specific widgets and injecting them into the big options-hash in highcharts-factory.svc.coffee. I think this could be refactored to merely merging the specific widgets' options hash with these default options, but this would take some time. For this pull request, I followed the current patterns of merging options.


iseessel:
@xaun I'm a bit confused a specific of your proposed solution. Overall, I'm a little confused about how we are populating the chartConfigOptions through the functions #template, #formatters, #todayMarker, and how that factors into your approach of having a single instance variable as @options.

So as it stands now, the highChart template points to a function that returns a hash populated based on settings and series. 

Since we intend to populate all the @options before we call render, I suggest that we populate @options in the constructor method, making the render method very lean. 

This change would require us to create a new instance of the highchartsFactory each time we give the chart new data, which I think is fine. 

Alternatively we could create an #updateOptions method, that updates the chart's @options (this was previously done in the #render method, but since we are changing the render methods to have no arguments, changing the data is no longer possible through the render method). 

`
class Chart
    constructor: (@id, @data = {}, @settings = {})->
      # Setup chart options
      templateOptions = @template(@settings.chartType)
      @options = angular.merge({}, templateOptions, @formatters(), @todayMarker())
      return

    render: ()->
      #It is faster to create a new stockChart than to update an existing one.
      @hc = Highcharts.stockChart(@id, @options)
      return @

    template: (chartType) ->
      settings = templates[chartType]
      settings.get(@data.series, @settings)
`

xaun:
@iseessel 

> how we are populating the chartConfigOptions through the functions #template, #formatters, #todayMarker, and how that factors into your approach of having a single instance variable as @options

Well, the current implementation is that the #template, #formatters & #todayMarkers are just returning an Object, which gets merged into the @options. This was the original pattern, then we thought it was nicer to leverage the highcharts `#update` method, to update small pieces of configurations without need to merge into @options, and after render. But since your discovery of performance issues, I was thinking we would go back to this, but adding chunks of "@options" with setter methods before #render. 

I like your idea to setup the chart options in the constructor, it doesn't look like it needs to be setup on render. 

Just a couple thing: 
- I think instead of sending a data object that only ever contains a series array, we can just give the series array directly and save it to @series.
- I think your example of the #template method needs some refining, and maybe @settings doesn't need to be an instance variable?

We give the #template method the chartType accessed from the @settings
```
templateOptions = @template(@settings.chartType)
```
But then the #template method internally access the @settings.
```
settings.get(@data.series, @settings)
```

I think @id, @series, @options should be the only instance variables, and you're right in that it would be more straight forward for all the setter methods to take arguments directly.

> the highChart template now points to a function that returns a hash populated based on the settings and series. Do we want to keep this pattern?

I'de like to not keep the pattern, ideally. All the options that use dynamic settings (`_.get(options, 'plotOptions.series.events', {})`) I think should be set after initalize, using setter methods. The template should just be boilerplate static options. What do you think?

It's just that big hash at the top with dynamic `_.get`'s, and now, conditional blocks, is getting a bit out of hand. What do you reckon is a more maintainable approach?

iseessel:
@xaun 

Along with your approach, I suggest we use short setter methods that take direct and specific arguments that populate the @options hash. Anything that is generated dynamically in @template will be refactored to a setter method that populates the @options hash.

iseessel:
@ouranos  Realized I was making PR onto 1.6 and not 1.6.9. This New Pull request made.

xaun:
Moved to PR #480

PR #470, opened 2018-02-05
Fix Chrome not launched on TravisCi (build still broken)

xaun:
@cesar-tonnoir I have tried to fix the build but cannot get it working. This at least fixes the chrome launching problems. 

I think the problems now are something to do with the dependency file loading. I have tried loading the core packages in a specific order, and excluding them from the Wiredep with no success. It is really a crappy system. 

Here is a glimpse of the current errors after my Chrome launcher fix:
screen shot 2018-02-05 at 12 42 05

In later attempts I manage to fix these exact problems by moving these deps (in karma.conf.js files) above the rest:
```
'bower_components/jasmine/lib/jasmine-core/jasmine.js',
'bower_components/jasmine-jquery/lib/jasmine-jquery.js',
```

But then other dep issues arose..

I cannot find a quick fix and would need more scope to look at a smarter solution (some people talking about requireJS, but setup doesn't look easy) for it unless you know what's wrong / have an idea?

I thought open these changes rather than delete the whole branch. Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,9 @@ language: node_js
 node_js:
   - "5.6.0"
 before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
+ - export CHROME_BIN=/usr/bin/google-chrome
+ - export DISPLAY=:99.0
+ - sh -e /etc/init.d/xvfb start
 before_script:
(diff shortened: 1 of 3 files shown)


xaun:
Merged into https://github.com/maestrano/impac-angular/pull/479

PR #480, opened 2018-02-17
[IMPAC-745] Fix Cash Projection Crash and Refactor Highcharts

iseessel:
@xaun
Implemented what we discussed. Its a pretty big pull request, so let me know if you have any questions. Let me know when you want me to squash ( would rather squash when ready to merge, as keeping track of my commits helps me).

I think this is an improvement over what was there originally. Only thing is its a bit confusing when @options should be populated in constructor vs in setter. All the dynamic options under the template, I put into setter methods, because they seemed specific to line charts.

### DIFF
--- a/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
+++ b/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
@@ -69,7 +69,7 @@ module.component('chartThreshold', {
 
     handleInvalidAlertAmount = ->
       toastr.error("Please choose a number one or greater.", 'Error')
-      ctrl.cancelCreateKpi()
+      ctrl.loading = false
 
     ctrl.saveKpi = ->
       return if ctrl.loading
--- a/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
+++ b/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
@@ -67,9 +67,6 @@ module.component('chartThreshold', {
       , 100)
       return
 
-    validAlertAmmount = (target) ->
-      ImpacKpisSvc.validateKpiTargets(target)
-
     handleInvalidAlertAmount = ->
       toastr.error("Please choose a number one or greater.", 'Error')
       ctrl.cancelCreateKpi()
@@ -82,13 +79,12 @@ module.component('chartThreshold', {
         "#{ctrl.kpiTargetMode}": parseFloat(ctrl.draftTarget.value)
       }]
 
-      if !validAlertAmmount(params.targets)
+      unless ImpacKpisSvc.validateKpiTargets(params.targets)
         return handleInvalidAlertAmount()
 
       promise = if ctrl.isEditingKpi
         ImpacKpisSvc.update(getKpi(), params, false).then(
           (kpi)->
-
             # Remove old threshold from chart
             ctrl.chart.removeThreshold(kpi.id)
             angular.extend(getKpi(), kpi)
@@ -103,7 +99,6 @@ module.component('chartThreshold', {
         )
       promise.then(
         (kpi)->
-
           ctrl.onComplete($event: { kpi: kpi }) if _.isFunction(ctrl.onComplete)
       ).finally(->
         ctrl.cancelCreateKpi()
@@ -133,7 +128,7 @@ module.component('chartThreshold', {
     onChartNotify = (chart)->
       ctrl.chart = chart
       return unless validateHistParameters()
-      ctrl.chart.options.chartOnClickCallbacks.push(onChartClick)
+      ctrl.chart.settings.chartOnClickCallbacks.push(onChartClick)
       _.each buildThresholdsFromKpis(), (threshold)->
         thresholdSerie = ctrl.chart.findThreshold(threshold.kpiId)
         thresholdSerie = ctrl.chart.addThreshold(threshold) unless thresholdSerie?
--- a/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
+++ b/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
@@ -128,7 +128,7 @@ module.component('chartThreshold', {
     onChartNotify = (chart)->
       ctrl.chart = chart
       return unless validateHistParameters()
-      ctrl.chart.settings.chartOnClickCallbacks.push(onChartClick)
+      ctrl.chart.addOnClickCallback(onChartClick)
       _.each buildThresholdsFromKpis(), (threshold)->
         thresholdSerie = ctrl.chart.findThreshold(threshold.kpiId)
         thresholdSerie = ctrl.chart.addThreshold(threshold) unless thresholdSerie?
--- a/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
+++ b/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
@@ -67,6 +67,13 @@ module.component('chartThreshold', {
       , 100)
       return
 
+    validAlertAmmount = (target) ->
+      ImpacKpisSvc.validateKpiTargets(target)
+
+    handleInvalidAlertAmount = ->
+      toastr.error("Please choose a number one or greater.", 'Error')
+      ctrl.cancelCreateKpi()
+
     ctrl.saveKpi = ->
       return if ctrl.loading
       ctrl.loading = true
@@ -74,10 +81,14 @@ module.component('chartThreshold', {
       params.targets[ctrl.kpi.watchables[0]] = [{
         "#{ctrl.kpiTargetMode}": parseFloat(ctrl.draftTarget.value)
       }]
-      return unless ImpacKpisSvc.validateKpiTargets(params.targets)
+
+      if !validAlertAmmount(params.targets)
+        return handleInvalidAlertAmount()
+
       promise = if ctrl.isEditingKpi
         ImpacKpisSvc.update(getKpi(), params, false).then(
           (kpi)->
+
             # Remove old threshold from chart
             ctrl.chart.removeThreshold(kpi.id)
             angular.extend(getKpi(), kpi)
@@ -92,6 +103,7 @@ module.component('chartThreshold', {
         )
       promise.then(
         (kpi)->
+
           ctrl.onComplete($event: { kpi: kpi }) if _.isFunction(ctrl.onComplete)
       ).finally(->
         ctrl.cancelCreateKpi()
--- a/src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
+++ b/src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
@@ -195,7 +195,6 @@ module.controller('WidgetAccountsCashProjectionCtrl', ($scope, $q, $filter, $tim
   w.format = ->
     _highChartOptions =
       chartType: 'line'
-      chartOnClickCallbacks: []
       currency: w.metadata.currency
       showToday: true
 
--- a/src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
+++ b/src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
@@ -194,9 +194,21 @@ module.controller('WidgetAccountsCashProjectionCtrl', ($scope, $q, $filter, $tim
   # Executed after the widget and its settings are initialised and ready
   w.format = ->
     # Instantiate and render chart
-    options =
-      chartType: 'line'
+    legendOptions =
+      legend:
+        useHTML: true
+        labelFormatter: legendFormatter
+
+    callBackOptions =
       chartOnClickCallbacks: []
+      plotOptions:
+        series:
+          events:
+            click: onClickBar
+            legendItemClick: onClickLegend
+
+    formattingOptions =
+      chartType: 'line'
       currency: w.metadata.currency
       showToday: true
       showLegend: true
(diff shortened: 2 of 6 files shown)


src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -43,30 +28,33 @@ angular
...
-          selected: (if _.get(xAxisOptions, 'min') then null else 0)
 
   todayUTC = moment().startOf('day').add(moment().utcOffset(), 'minutes')
 
   class Chart
-    constructor: (@id, @data = {}, @options = {})->
-      @_template = templates[@options.chartType]
+    constructor: (@id, @series = {}, @settings = {})->
+      @_template = templates[@settings.chartType]
+      @populateChartOptions(@series, @settings)
```
xaun: Not a big fan of how this method is public, seems more like a private method, and I think everything that happens in that method could just be within the constructor block.

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -43,30 +28,33 @@ angular
...
-      if _.isEmpty(@hc)
-        @hc = Highcharts.stockChart(@id, chartConfig)
-      else
-        @hc.update(chartConfig)
+    populateChartOptions: (series, settings) ->
+      # Every chart will have these options. The rest of the options are created in the setter methods.
+      template = @_template
+      formatters = @formatters(settings.currency)
+      todayMarker = @todayMarker(settings)
+      series = @populateSeries(series)
```
xaun: Prefer `series = { series: series }`, I dont think we need a `populateSeries` public instance method here

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -43,30 +28,33 @@ angular
...
-      @data = data if _.isObject(data)
-      angular.extend(@options, options)
-      chartConfig = angular.merge({}, @template(), @formatters(), @todayMarker())
-      if _.isEmpty(@hc)
-        @hc = Highcharts.stockChart(@id, chartConfig)
-      else
-        @hc.update(chartConfig)
+    populateChartOptions: (series, settings) ->
+      # Every chart will have these options. The rest of the options are created in the setter methods.
+      template = @_template
```
xaun: If you move the contents of this `populateChartOptions` method to just the constructor block, as @_template is not being accessed anywhere, we can remove it from the `this` scope, and just use 
```
template = templates[@settings.chartType]
angular.merge({}, templates ... ... . .) 
```

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -86,11 +74,11 @@ angular
             name = _.startCase _.trim name.toLowerCase().replace(/\s*projected\s*/, ' ')
           "<strong>#{date}</strong><br>#{name}: #{amount}"
 
-    todayMarker: ->
-      return {} unless @options.showToday
+    todayMarker: (settings) ->
```
xaun: `todayMarker: (showToday, markerColor = 'rgba(0, 85, 255, 0.2)')` ?

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -130,22 +118,48 @@ angular
...
-        useHTML: useHTML
-        labelFormatter: formatterCallback
-      })
+    addCustomLegend: (labelFormatter, useHTML = true, showLegend = true) ->
+      legend =
+        legend:
+          labelFormatter: labelFormatter
+          useHTML: useHTML
+          enabled: showLegend
+      @options = angular.merge(@options, legend)
```
xaun: The `angular.merge` method is destructive. The object in argument 1 will have the following object args merged into it. So @options will be mutated. If you want non destructive merging, pass an empty hash as the first argument like so:
```
# Note that @options will not contain attributes from someOtherHash
myNewHash = angular.merge({}, @options, someOtherHash)
```
So here you could simply write:
```
angular.merge(@options, legend)
```

src/components/widgets/accounts-class-comparison/accounts-class-comparison.directive.coffee
```diff
@@ -35,7 +35,6 @@ module.controller('WidgetAccountsClassComparisonCtrl', ($scope, $q, $filter, Cha
   w.initContext = ->
     $scope.isDataFound = angular.isDefined(w.content) && !_.isEmpty(w.content.summary) && !_.isEmpty(w.content.companies)
     if $scope.isDataFound
-      $scope.timePeriodInfoParams.histParams = w.metadata && w.metadata.hist_parameters
```
xaun: Why is this removed?

src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
```diff
@@ -173,6 +184,14 @@ module.controller('WidgetAccountsCashProjectionCtrl', ($scope, $q, $filter, $tim
       continue if s.userOptions.linkedTo != series.name
       if series.visible then s.hide() else s.show()
 
+  addCustomHighChartOptions = () ->
+    highChartsFactory = new HighchartsFactory($scope.chartId(), w.content.chart.series, _highChartOptions)
+    highChartsFactory.addCustomLegend(legendFormatter)
+    highChartsFactory.addSeriesEvent('click', onClickBar)
+    highChartsFactory.addSeriesEvent('legendItemClick', onClickLegend)
+    highChartsFactory.addXAxisOptions(_highChartOptions.withZooming)
```
xaun: I like the setters, means widgets can really configure the charts with a nice api. 

Thought It feels weird giving the factory the full `_highChartOptions` on init, then accessing one of the attributes for the `addXAxisOptions` method. I wouldn't shy away from using this syntax:

```coffeescript
highChartsFactory.addXAxisOptions({
    defaults: w.metadata.xAxis
    callback: onZoom
})
```

It means I you dont have to go scrolling up and down the file looking for the options hash, keeping things together. I think we should move the whole options to this if possible, see my comment at the `_highChartOptions` declarations.

src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
```diff
@@ -33,6 +33,17 @@ module.controller('WidgetAccountsCashProjectionCtrl', ($scope, $q, $filter, $tim
   $scope.chartId = ->
     "cashProjectionChart-#{w.id}"
 
+  # == Setup Highchart Options ============================================================================
+
+  _highChartOptions = Object.freeze
+    chartType: 'line'
+    chartOnClickCallbacks: []
+    currency: w.metadata.currency
```
xaun: As the variable `_highChartsOptions` is being defined in the controller function level of this directive, it will be evaluation as soon as the function is invoked internally by angular. This means, I am quite confident, w.metadata will be undefined at that exact point in time. 

This is essentially being decared as a constant, but as it has dynamic data inside it, it needs to either be `$onInit` (still fakey with asyc data), or better `$onChanges`, or even better (less code), re-evaluated when `w.format` is called.

src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
```diff
@@ -67,14 +67,21 @@ module.component('chartThreshold', {
...
     ctrl.saveKpi = ->
       return if ctrl.loading
       ctrl.loading = true
       params = targets: {}, metadata: {}
       params.targets[ctrl.kpi.watchables[0]] = [{
         "#{ctrl.kpiTargetMode}": parseFloat(ctrl.draftTarget.value)
       }]
-      return unless ImpacKpisSvc.validateKpiTargets(params.targets)
+
+      if !ImpacKpisSvc.validateKpiTargets(params.targets)
```
xaun: prefer `unless` to `if !` with coffee

src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
```diff
@@ -67,14 +67,21 @@ module.component('chartThreshold', {
       , 100)
       return
 
+    handleInvalidAlertAmount = ->
+      toastr.error("Please choose a number one or greater.", 'Error')
+      ctrl.cancelCreateKpi()
```
xaun: Is this new functionality? Do we want to `cancelCreateKpi` if I enter a wrong number?

src/services/highcharts-factory/highcharts-factory.svc.coffee
iseessel: Sure.

src/components/widgets/accounts-class-comparison/accounts-class-comparison.directive.coffee
iseessel: Oops most likely a rebasing problem.

src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
iseessel: Sure.

src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
iseessel: Yes we do, currently when we enter a wrong number, the update box will keep on spinning.

src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
iseessel: Good catch. Will put back into w.format.

src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
iseessel: Good point. What exactly do you mean by "we should move the whole options to this if possible."

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -43,30 +28,33 @@ angular
...
             { type: 'all', text: 'All' }
           ]
-          selected: (if _.get(xAxisOptions, 'min') then null else 0)
 
   todayUTC = moment().startOf('day').add(moment().utcOffset(), 'minutes')
 
   class Chart
-    constructor: (@id, @data = {}, @options = {})->
-      @_template = templates[@options.chartType]
+    constructor: (@id, @series = {}, @settings = {})->
```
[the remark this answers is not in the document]
iseessel: Yes, as can id.

src/services/highcharts-factory/highcharts-factory.svc.coffee
iseessel: Sure thing, can just move code over.

src/services/highcharts-factory/highcharts-factory.svc.coffee
iseessel: Yeah that works.

src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee
xaun: @iseessel never-mind about that comment actually

src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
xaun: @iseessel I had a quick look, the problem is the `ctrl.saveKpi` method sets `ctrl.loading = true`, but then returns without setting the `ctrl.loading` back to `false` unless `ImpacKpisSvc.validateKpiTargets(params.targets)`.

The quick solution to use a toastr to notify the user is fine I guess, though ngMessages txt would probs be better. But instead of calling `ctrl.cancelCreateKpi()` in your `handleInvalidAlertAmount` method, can we just try setting `ctrl.loading` to `false`? I think it would be a better ux.

src/services/highcharts-factory/highcharts-factory.svc.coffee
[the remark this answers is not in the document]
iseessel: EDIT: We can't remove settings from this, as it is called in #onChartNotify in `chart-threshold.component.coffee`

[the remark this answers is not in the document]
xaun: Ah ok, do you think we should rename `@options` to `@hcOptions` or `@highchartsOptions` for clarity between options & settings?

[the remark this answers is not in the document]
iseessel: Yeah probably a good idea.

src/components/widgets-common/chart-threshold/chart-threshold.component.coffee
iseessel: agreed

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -4,37 +4,22 @@ angular
...
-        zoomingOptions = _.get(options, 'withZooming')
-        xAxisOptions = if zoomingOptions?
-          {
-            events:
-              setExtremes: zoomingOptions.callback
-            max: _.get(zoomingOptions.defaults, 'max')
-            min: _.get(zoomingOptions.defaults, 'min')
-          }
-
         chart:
```
xaun: @iseessel there is inconsistent indentation here. The `line` object properties are indented 4 spaces, and the rest of the file is 2 spaces.

xaun:
@iseessel if you could squash all commits like "refactoring", "debugging", "fix rebase problem",  "comments", "take out debugging". Commits like "Separated multiple update calls into one create call" can remain. Not sure why you need to keep "take out debuggers" commits during your  I development process? I suggest always using `git add -p` to stage your changes in hunks, giving you the chance to audit your work in pieces before committing, therefor not needing to remove debuggers later, and squash "remove debuggers" commits.

iseessel:
@xaun Found a bug where an onclick callback did not work. Was quite a doozy to find the problem. 

It boiled down to a pointer program. You need the chartOnClickCallbacks to point to settings.onClickCallbacks, because in `chart-threshold-component.coffee' you push another callback into this array.

iseessel:
@xaun So add manual transaction does not work in the cash projection widget, getting a 400 error from the server -- this is an existing error (tested it pre changes as well) and I suggest putting it in a new issue.

xaun:
@iseessel I think there may be a better way that follows the new concept of setting options then rendering / re-rendering. 

If you look [here](https://github.com/iseessel/impac-angular/blob/d66bf5f814dea5c0a840da0682d8e8319ceeed9f/src/components/widgets-common/chart-threshold/chart-threshold.component.coffee#L44) you can see this is where the chart-threshold.component is calling `onChartNotify`. 

`ctrl.chartPromise` is a component binding given by the parent widget.

If we take a look [here](https://github.com/iseessel/impac-angular/blob/d66bf5f814dea5c0a840da0682d8e8319ceeed9f/src/components/widgets/accounts-cash-projection/accounts-cash-projection.directive.coffee#L214) at the cash-projection.directive, (`$scope.chartDeferred` is bound to chart-threshold.component as `chartPromise`), the notify is being invoked after render. I think if we simply invoke `$scope.chartDeferred.notify($scope.chart)` before calling `$scope.chart.render()`, it will work.

PR #482, opened 2018-02-26
[IMPAC-764] Invoices List Widget v2

paulstuartparker:


src/components/widgets/accounts-invoices-list/accounts-invoices-list.directive.coffee
```diff
@@ -0,0 +1,66 @@
...
+
+  # Widget specific methods
+
+  $scope.trxList.show = ->
+    $scope.trxList.display = true
+
+  $scope.trxList.hide = ->
+    $scope.trxList.display = false
+
+  $scope.trxList.fetch = (currentPage = 1) ->
```
cesar-tonnoir: Quick question: do you think this could be moved to the transactions-list component itself?

paulstuartparker: I don't think so, because sometimes the show or hide method may require extra logic. For instance in the accounts-cash-projection widget: 
```
  $scope.trxList.show = ->
    $scope.trxList.display = true

  $scope.trxList.hide = ->
    $scope.trxList.display = false
    if $scope.trxList.updated
      ImpacWidgetsSvc.show(w).then(-> $scope.trxList.updated = false)
```

src/components/widgets/accounts-invoices-list/accounts-invoices-list.directive.coffee
```diff
@@ -4,11 +4,10 @@ module.controller('WidgetAccountsInvoicesListCtrl', ($scope, $q, ImpacRoutes, Bo
   # == Context and Helpers ========================================================================
   w = $scope.widget
   bolts = ImpacRoutes.bolts()
-  bolt_path = _.find(bolts, {name: 'finance', provider: 'maestrano'}).path
+  bolt_path = w.metadata.bolt_path
```
cesar-tonnoir: Are you sure this would still work when you add the widget to the dashboard the first time? Not sure the `metadata` would be set already in this case?...
What was wrong with the previous solution?

src/components/widgets/accounts-invoices-list/accounts-invoices-list.directive.coffee
```diff
@@ -4,11 +4,10 @@ module.controller('WidgetAccountsInvoicesListCtrl', ($scope, $q, ImpacRoutes, Bo
...
 
   $scope.trxList = {
     display: false,
-    resources: 'invoices',
```
cesar-tonnoir: Why changing that?

src/components/widgets/accounts-invoices-list/accounts-invoices-list.directive.coffee
```diff
@@ -52,6 +51,7 @@ module.controller('WidgetAccountsInvoicesListCtrl', ($scope, $q, ImpacRoutes, Bo
   # --------------------------------------
   w.initContext = ->
     $scope.isDataFound = w.content?
+    $scope.trxList.resources = w.content.invoice_endpoint
```
cesar-tonnoir: You should not rely on what the api will return to tell your frontend component that it will receive invoices / the previous way seemed fine to me

paulstuartparker:
@cesar-tonnoir I agree with you, I like the previous solution too.  I guess I misunderstood what you were asking.  Should I revert the changes?

PR #484, opened 2018-03-05
[IMPAC-793] Dynamically load generic layout templates

xaun:
For legacy widgets, behaviour is the same as before. If the widget template doesn't exist, the widget is not available in the widgets list. 

For bolt widgets, widgets are shown in the widgets list unless blacklisted. When added, checks for a custom template in dir `widgets-layouts/custom`, if no custom template is defined, attempts to load a generic layouts template from e.g dir `widgets-layouts/{{layoutName}}`, if no layout template is found, removes the loading widget from the dashboard, sends a toastr alert, and $log's details in the console. 

**Cash Balance & Cash Projection widgets files have been moved to `widgets-layouts/custom` dir**

- [ ] Dependency versions specified in Changelog (Mnohub: maestrano/maestrano-hub#915, Mnoe: maestrano/mno-enterprise#665


src/components/widget/widget.directive.coffee
```diff
@@ -75,16 +75,22 @@ module.directive('impacWidget', ($templateCache, ImpacUtilities, ImpacWidgetsSvc
...
-      scope.pdfMode = false
-      scope.cssClass = ImpacUtilities.fetchWidgetCssClass(scope.widget)
-      scope.templatePath = ImpacUtilities.fetchWidgetTemplatePath(scope.widget)
+      onInit = ->
+        scope.showInfoPanel = false
+        scope.pdfMode = false
+        scope.widget.isLoading = true
+        scope.widget.settings = []
+        # Unused so far -->
+        scope.widget.hasEditAbility = true
```
cesar-tonnoir: I think this is old code that has been removed with the implementation of ACLs (v1.7.0) => can you remove plz?

src/components/widget/widget.directive.coffee
```diff
@@ -106,6 +112,14 @@ module.directive('impacWidget', ($templateCache, ImpacUtilities, ImpacWidgetsSvc
         ImpacWidgetsSvc.delete(scope.widget)
         .then(null, (e) -> scope.widget.errors = ImpacUtilities.processRailsError(e))
 
+      _handleNoTemplateFound = ->
```
cesar-tonnoir: :+1:

src/components/widget/widget.directive.coffee
xaun: Sure 👍

PR #487, opened 2018-03-09
Feature/lmi insights

MarcoBgn:
- Initial WIP changes for the LMI feature


manu-d:
@cesar-tonnoir 
Can you have a look at this PR, so that I can open all my PR against maestrano's repo and not Marco's?
Thanks

src/components/widgets/sales-average-purchase-size/sales-average-purchase-size.directive.coffee
```diff
@@ -0,0 +1,88 @@
...
+              if (this.chart.rangeSelector.options.selected >= 3)
+                moment.utc(this.value).format('MMM YYYY')
+              else
+                moment.utc(this.value).format('DD MMM')
+        yAxisLabels =
+          labels:
+            formatter: ->
+              $filter('mnoCurrency')(this.value, currency, false)
+        xAxis: angular.merge([w.content.chart.xAxis[0]], [xAxisLabels])
+        yAxis: angular.merge([w.content.chart.yAxis[0]], [yAxisLabels])
```
xaun: The yAxisLabels are the same as default (right?), for overriding the formatters method, I suggest leveraging the formatters method's returned defaults by savings them to a variable in the outer scope.

Also, rather than having to merge the new v2 bolt layouts response options (xAxis, yAxis, tooltip etc) for every single custom widget template, the highcharts factory should be extended to use these by default.

 Then you could do this sort of thing:
```coffeescript
defaultFormatters = $scope.chart.formatters()
$scope.chart.formatters = ->
  defaultFormatters.labels.formatter = -> 
    formatString = if this.chart.rangeSelector.options.selected >= 3 then 'MMM YYY' else 'DD MMM'
    moment.utc(this.value).format(formatString)
  defaultFormatters.rangeSelector = { selected: 5 }
  defaultFormatters
```

EDIT: so I can see now that you have extended the highcharts formatter, so would simply be saving the `defaultFormatters` be enough to not have to do the `angular.merge([w.content.chart.xAxis[0] .. )` ? Seems like you are giving the formatter the w.content.chart.xAxis in initialization, but then having to re-provide it with it to merge in the custom opts.

src/components/widgets/sales-customers-engagement/sales-customers-engagement.tmpl.html
```diff
@@ -0,0 +1,30 @@
...
+    <div class="bottom-buttons" align="right">
+      <button class="btn btn-default" ng-click="initSettings()">Cancel</button>
+      <button class="btn btn-warning" ng-click="updateSettings()">Save</button>
+    </div>
+  </div>
+
+  <!-- Content Panel -->
+  <div ng-hide="widget.isEditMode">
+    <!-- Data found -->
+    <div ng-show="isChartDisplayed" class="">
```
xaun: Does `isChartDisplayed` ever get set false?

src/services/highcharts-theme/highcharts-theme.svc.coffee
```diff
@@ -0,0 +1,32 @@
+angular
+.module('impac.services.highcharts-theme', [])
+.service('HighchartsThemeService', (ImpacTheming) ->
```
xaun: Looks more like something that should live in an angular `.run()` function?

src/services/highcharts-factory/highcharts-factory.svc.coffee
```diff
@@ -63,7 +69,7 @@ angular
       return @
 
     template: ->
-      @_template.get(@data.series, @options)
+      @_template.get(@data.series, @options, @data.xAxis, @data.yAxis, @data.plotOptions)
```
xaun: Ok for now, @cesar-tonnoir will latest 1.6 be back-merged into lmi? If so there will be lots of conflicts as this whole factory has been reworked (#480) to simplify providing options.

cesar-tonnoir: @xaun I don't think that will be necessary: the LMI widgets will use the figure+chart generic template (will be developed in 1.8 with your implementation of generic templates loading), so ultimately we shouldn't need that many customisations

src/components/widgets/sales-customers-engagement/sales-customers-engagement.tmpl.html
MarcoBgn: @manu-d as discussed could you check this?

src/components/widgets/sales-average-purchase-size/sales-average-purchase-size.directive.coffee
MarcoBgn: @xaun I am not completely sure about this as there seem to be slight differences in the overridden formatters

src/services/highcharts-theme/highcharts-theme.svc.coffee
MarcoBgn: I am having issues with `Highcharts.theme =` where Highcharts is not defined yet. Not being a service (i.e. no HighcartsProvider), I am not sure this can live in a run() function

src/components/widgets/sales-customers-engagement/sales-customers-engagement.tmpl.html
xaun: @manu-d ?

src/services/highcharts-factory/highcharts-factory.svc.coffee
xaun: @cesar-tonnoir when you say "will be developed", is that different to merging this branch into upstream/1.x ? Seems there is quite a bit of HighchartsFactory#formatter customisation (of which I am not too happy with the resulting code), will the formatter stuff be discarded for generic templates? Or will it live separately somehow?

src/components/widgets/sales-average-purchase-size/sales-average-purchase-size.directive.coffee
xaun: @MarcoBgn Im not completely across the intricacies of your customisations, so my solution isn't comprehensive for your use-case. I am just showing you a way you can store the existing defaults, then modify the object and reassign the formatters method. 

The main thing is, this method is massive in the widget controller, it redefines **some** values that are already defined in the default, and looks very similar across all of the new widgets. So for my review, I request that we think of a solution to improve this. If this is not in the scope of this body of work, or it needs to be quick & dirty merged or something, that's another story :)

PR #489, opened 2018-03-12
[IMPAC-739] Add new figure layout

manu-d:


src/components/widgets-layouts/figure/figure.directive.coffee
```diff
@@ -0,0 +1,22 @@
...
+    ChartFormatterSvc.getColor(index)
+
+)
+
+module.directive('impacFigureLayout', ($templateCache) ->
+  return {
+    restrict: 'E',
+    scope: {
+      content: '='
+      timePeriodEnabled: '=?'
```
xaun: `timePeriodEnabled`, `selectorList`, and `selectorUid` isn't used anywhere.

src/components/widgets-layouts/figure/figure.less
```diff
@@ -0,0 +1,10 @@
+.analytics .layouts.figure {
```
xaun: No styles are being applied to indicate this is a column or scrollable as specified in the JIRA. 
In screenshot below I have added entries and they just extend the height of the widget. 
screen shot 2018-03-19 at 11 23 56

PR #490, opened 2018-03-12
[IMPAC-698] Add widget average spend per customer

manu-d:


xaun:
Also @manu-d, the JIRA says to load a generic Highcharts template?

PR #504, opened 2018-03-21
[IMPAC-811] Add trend in cash projection widget

manu-d:


src/components/common/trends-add/trends-add.component.coffee
```diff
@@ -0,0 +1,51 @@
...
+    ctrl.isValid = ->
+      !_.isEmpty(ctrl.trend.name) &&
+      ctrl.trend.rate > 0 &&
+      ctrl.trend.untilDate? || ctrl.trend.period == "Once"
+
+    ctrl.createTrend = ->
+      ctrl.onHide()
+      ctrl.onCreateTrend({ trend: ctrl.trend })
+
+    $scope.$watch '$ctrl.datePicker.opened', (newVal, oldVal) ->
```
xaun: Generally speaking, there shouldn't be too many cases where you actually need to use a `$watch`. 

Correct me if Im wrong, but couldn't your `datePicker.toggle` method just apply the logic that this $watch callback is? As that seems to be what is setting `datePicker.opened` anyway.

src/components/common/trends-add/trends-add.tmpl.html
```diff
@@ -0,0 +1,34 @@
...
+      <select ng-model="$ctrl.trend.period">
+        <option>Once</option>
+        <option>Daily</option>
+        <option>Weekly</option>
+        <option>Monthly</option>
+        <option>Annually</option>
+      </select>
+
+      <fieldset class="btn-group" ng-disbaled="$ctrl.isPeriodDisabled()">
+        <input type="radio" ng-model="$ctrl.trend.untilDate" name="until" checked="true" ng-value=-1> Forever<br>
```
xaun: The `ng-value` is missing quotations around the value on this input and the input below this line. Are you sure its working / needed?

PR #505, opened 2018-03-26
[IMPAC-815] - Forecast Payroll

agranado2k:
Enable or disable Forecast Payroll projection


src/components/widgets-settings/forecast-payroll/forecast-payroll.tmpl.html
```diff
@@ -0,0 +1,10 @@
+<div class="settings forecast-payroll">
+  <h5 translate>impac.widget.settings.forecast_payroll.title</h5>
+
+  <div class="widget-lines-container">
+    <div class="widget-line">
+      Aggregate
```
cesar-tonnoir: Should be in a locale.
Can you use something like: "Include payroll costs into cash projection"?

agranado2k: done

PR #510, opened 2018-04-05
[IMPAC-792] Generic template for table + highstocks layouts

xaun:
@cesar-tonnoir take a look at how I am using the `HighchartsFactory`, and the small changes I made to it to not have to pass all the options/arguments it requires to build a factory object. 

It doesn't need to be a factory any more, as the static templates system is useless with the layouts. We could just use a service which has methods that return Highcharts options as plain objects. The only options we need built on the FE are options with events, and the add threshold functionality (which is done post-render anyway). 

Let me know if you think I should do this in this PR, or another, and whether it should be in v1.8 or v1.9, or if you think otherwise about the refactor? Thanks

Here is the end result of the profit_and_loss widget loaded with the generic "table-highstocks" template:
screen shot 2018-04-07 at 14 05 29


xaun:
bolt `preferred_layout` PR: https://github.com/maestrano/impac-finance-bolt/pull/117

PR #413, opened 2017-09-27
[IMPAC-639] Threshold currency

xaun:
Rebased this feature on-top of 1.6.1. The last commit contains fixes & refactors required due the the latest changes in 1.6.1. All previous commits were reviewed here: https://github.com/maestrano/impac-angular/pull/400.

PR #511, opened 2018-04-11
Archive/threshold currency

cesar-tonnoir:
Replaces #413


xaun:
Replaced by https://github.com/maestrano/impac-angular/pull/514 (which rebases these changes on-top of 1.7)

PR #522, opened 2018-05-02
[IMPAC-868] Implemented improvements on invoices list widget

abdullahshafiq63:


src/components/widgets/accounts-invoices-list/accounts-invoices-list.directive.coffee
```diff
@@ -53,6 +61,17 @@ module.controller('WidgetAccountsInvoicesListCtrl', ($scope, $q, ImpacRoutes, Bo
     $scope.trxList.resources = resourcesType
     $scope.trxList.fetch()
 
+  $scope.trxList.changeOverdueFilter = (overdueFilter) ->
```
cesar-tonnoir: Shouldn't it be in the "transactions-list" component scope?
I mean, if I want to use the transactions-list component in another widget, the behaviour will be the same isn't it?
Same comment for `changeQuery`

abdullahshafiq63: @cesar-tonnoir both 'changeOverdueFilter' and 'changeQuery' make use of the fetch function hence its written in parent scope.

Both 'invoices-list' and 'cash-projection' are responsible for fetching transactions list and passing this down to 'transaction-list' component, which simply displays them in a table.

In order to perform some search or filter we have to notify the parent in order to fetch transactions again and pass them to transaction-list component to display the updated results.

I followed the already implemented flow of 'changeResourcesType' i.e 'see money in/ see money out' filter.

cesar-tonnoir: ok, fair enough / we may need to revise this logic then but that's not in the scope of this task

PR #527, opened 2018-05-10
[IMPAC-850] Add confirm modal for delete transactions & fixes

xaun:
- Refactored the existing delete-widget & delete-schedule into one component, and now also using it for delete-transaction. This makes all the widget modal interacts streamline. I have used transclusion to give the flexibility to use this component for a wider array of applications (based on @agranado2k saying the schedule modal needs to be extended soon).
       screen shot 2018-05-10 at 16 12 17
- Fix transactions-list `deleteTransactionsGroup` method breaking the promise chain
- Fix no cursor pointer for schedule transaction action button 

@cesar-tonnoir I think this should go in with 1.7.4, as there is a confirm modal for delete schedule transactions, but no confirm modal for deleting transactions, if you disagree, I can apply the two fixes separately. Thanks


src/components/common/transactions-list/transactions-list.component.coffee
```diff
@@ -141,21 +165,23 @@ module.component('transactionsList', {
...
       # TODO: should be accessible by recurring transactions only
-      return unless trx.recurring_parent
-      ctrl.onDeleteParentTransaction({ resourcesType: ctrl.resourcesType, trxId: trx.recurring_parent })
-      # Remove all children transactions and parent transaction if it is a forecast
-      _.remove(ctrl.currentAttributes.transactions, (trxInList) ->
-        (trxInList.recurring_parent == trx.recurring_parent) || (trxInList.id == trx.recurring_parent && trxInList.status == 'FORECAST')
+      return $q.when(null) unless trx.recurring_parent
+      args = { resourcesType: ctrl.resourcesType, trxId: trx.recurring_parent }
+      ctrl.onDeleteParentTransaction(args).finally(->
+        # Remove all children transactions and parent transaction if it is a forecast
```
cesar-tonnoir: I wouldn't add a `finally` there: we don't need to wait for the API response to remove the elements from the list

PR #529, opened 2018-05-14
Refactor transactions-list component

agranado2k:
- Redesign delete transaction and schedule:
-- it seems more calendar action where you have option to delete all occurrences or only one
- BugFix to create schedule from real transaction


agranado2k:
These stuffs were added in another PR
https://github.com/maestrano/impac-angular/pull/559

PR #537, opened 2018-05-17
[IMPAC-837][IMPAC-838] Add grouping of trends, and visualization when adding trends

manu-d:


xaun:
Replace by #542 and #543

PR #540, opened 2018-05-21
[MARLIN-123] Feature/rec engine 1.0.0

MarcoBgn:
- Add settings tab allowing Users to choose the data source (i.e primary, Xero, Manual)
- Add field to Widget service allowing Settings Options to be passed


src/components/widgets-settings/source-selector/source-selector.tmpl.html
```diff
@@ -0,0 +1,10 @@
+<div class="settings source-selector">
+  <h5 translate>impac.widget.settings.source-selector.select_source</h5>
+
+  <div class="widget-lines-container">
+    <div class="widget-line" ng-repeat="app in organizationApps">
+      {{app.label}}
```
xaun: Best to wrap this text in a more semantic element, like `<p>`, `<span>`, or `<label>` (https://www.w3.org/TR/html5/grouping-content.html#the-div-element)

src/components/widgets-settings/source-selector/source-selector.directive.coffee
```diff
@@ -0,0 +1,80 @@
...
+      parentWidget: '='
+      deferred: '='
+    },
+    template: $templateCache.get('widgets-settings/source-selector.tmpl.html'),
+
+    link: (scope) ->
+      w = scope.parentWidget
+
+      # Will there be a multi-org use case?
+      scope.mode ||= 'single'
```
xaun: There is no binding for `mode`, so it will never be previously set, is this intentional is it's a todo?

MarcoBgn: Yes, at the moment only single organization is being managed. I have added a `TODO` to make it more explicit

PR #559, opened 2018-06-13
Improve Transaction-list component

agranado2k:
- merge the options to delete transactions (only one and all recurring)

### DIFF
--- a/src/components/common/transactions-list/transactions-list.component.coffee
+++ b/src/components/common/transactions-list/transactions-list.component.coffee
@@ -210,6 +210,8 @@ module.component('transactionsList', {
         trxGroupId = trx.id
         trx.recurring = false
       else
+        # trx.recurring_parent is recurring_parent's id
+        # TODO: rename recurring_parent to recurring_parent_id when the resource in Bolt change id
         return unless trx.recurring_parent
         trxGroupId = trx.recurring_parent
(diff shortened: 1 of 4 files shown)


src/components/common/transactions-list/transactions-list.component.coffee
```diff
@@ -206,7 +206,14 @@ module.component('transactionsList', {
 
     # Remove all children and update parent
     deleteTransactionsGroup = (trx) ->
-      trxGroupId = if trx.recurring then trx.id else trx.recurring_parent
+      if trx.recurring
+        trxGroupId = trx.id
+        trx.recurring = false
+      else
+        return unless trx.recurring_parent
+        trxGroupId = trx.recurring_parent
```
MarcoBgn: Shouldn't be `trx.recurring_parent.id` here?

MarcoBgn (commented):
@agranado2k just a comment about the `trxGroupId`

src/components/common/transactions-list/transactions-list.component.coffee
agranado2k: Actually no because `trx.recurring_parent` is already id... the name is not good... should be `trx.recurring_parent_id`. I agree.. but the resource in Bolt is like it... :/

MarcoBgn: Right, can you add a comment to avoid future headaches? :)

MarcoBgn (approved):

PR #560, opened 2018-06-20
[IMPAC-905] Resolved bug with filtering past due invoices in invoices…

abdullahshafiq63:
… list widget

### DIFF
--- a/src/components/widgets-layouts/custom/accounts-invoices-list/accounts-invoices-list.directive.coffee
+++ b/src/components/widgets-layouts/custom/accounts-invoices-list/accounts-invoices-list.directive.coffee
@@ -68,8 +68,12 @@ module.controller('WidgetAccountsInvoicesListCtrl', ($scope, $q, ImpacRoutes, Bo
   $scope.trxList.changeOverdueFilter = (overdueFilter) ->
     return if overdueFilter == $scope.trxList.overdue
     $scope.trxList.overdue = overdueFilter
+    todays_date = new Date()
+    formated_date = [todays_date.getFullYear(), (todays_date.getMonth() + 1), todays_date.getDate()].join('-')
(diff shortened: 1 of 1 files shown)


src/components/widgets-layouts/custom/accounts-invoices-list/accounts-invoices-list.directive.coffee
```diff
@@ -68,8 +68,12 @@ module.controller('WidgetAccountsInvoicesListCtrl', ($scope, $q, ImpacRoutes, Bo
   $scope.trxList.changeOverdueFilter = (overdueFilter) ->
     return if overdueFilter == $scope.trxList.overdue
     $scope.trxList.overdue = overdueFilter
+    todays_date = new Date()
+    formated_date = [todays_date.getFullYear(), (todays_date.getMonth() + 1), todays_date.getDate()].join('-')
```
agranado2k: What do you think to use momentjs here? We use it a lot and easy you could do
`todays_date = moment().format('YYYY-MM-DD')`.

src/components/widgets-layouts/custom/accounts-invoices-list/accounts-invoices-list.directive.coffee
```diff
@@ -68,8 +68,12 @@ module.controller('WidgetAccountsInvoicesListCtrl', ($scope, $q, ImpacRoutes, Bo
...
     $scope.trxList.params.filter.balance = 'gt 0' if overdueFilter == 'overdue'
+    $scope.trxList.params.filter.due_date = ['lte', formated_date].join(' ') if overdueFilter == 'overdue'
```
agranado2k: What do you think about use string interpolation here? something like `"lte #{todays_date}"`

agranado2k (commented):
Hi Abdullah, I have just two small comments. What do you think?

agranado2k (approved):
LGFM

PR #562, opened 2018-06-22
[IMPAC-865] BugFix - metadata values were changed when the widget was

agranado2k:
load because in the intiazilaztion of the settings for averages.

### DIFF
--- a/src/components/widgets-settings/averages/averages.directive.coffee
+++ b/src/components/widgets-settings/averages/averages.directive.coffee
@@ -56,8 +56,9 @@ module.controller('SettingAveragesCtrl', ($scope, ImpacDashboardsSvc, GRANULARIT
         currentDashboard = config.currentDashboard
         averages = _.get(w, 'metadata.averages')
 
+        console.log('init')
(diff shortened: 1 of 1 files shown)


src/components/widgets-settings/averages/averages.directive.coffee
```diff
@@ -56,8 +56,9 @@ module.controller('SettingAveragesCtrl', ($scope, ImpacDashboardsSvc, GRANULARIT
         currentDashboard = config.currentDashboard
         averages = _.get(w, 'metadata.averages')
 
+        console.log('init')
```
MarcoBgn: Console.log forgotten here

MarcoBgn (changes requested):
@agranado2k looks good, just some logging to be removed

src/components/widgets-settings/averages/averages.directive.coffee
agranado2k: thanks

MarcoBgn (approved):

PR #565, opened 2018-11-29
[KAPP-175]  Allow runtime services configuration

ouranos:
This allow the admin panel to switch Impac configuration between staff
dashboards and designer mode in different parts of the application.

__Note:__ I've used the same method  as the `provider.configure` which is why we have `extend` and `merge`
Ok to review but I'd wait for extensive testing and QA before merging


xaun (approved):

cesar-tonnoir (approved):
LGTM - Ok, let's test in UAT and merge/tag if all good

PR #566, opened 2019-01-10
Fix Demo Data

ouranos:
* Fix Demo Data params that was incorrectly set
* Hide the "Add App" link when no message is set for this link

### Current Behaviour with visible Link

image

image

### Hidden Link

```ruby
dataNotFoundConfig:
      content:
        linkMessage: null
```

image
image

### DIFF
--- a/src/components/common/data-not-found/data-not-found.tmpl.html
+++ b/src/components/common/data-not-found/data-not-found.tmpl.html
@@ -3,11 +3,13 @@
   <div class="message" ng-show="messageVisible" >
     <div class="title" translate>{{ needsCustomMessage ? content.liveBTitle : content.title }}</div>
     <p translate>{{ needsCustomMessage ? content.liveBMessage : content.title }}</p>
-    <button class="btn btn-xs btn-default" ng-click="goToMarketplace()" translate>{{ content.linkMessage }}</button>
+    <button class="btn btn-xs btn-default" ng-click="goToMarketplace()" ng-if="content.linkMessage" translate>{{ content.linkMessage }}</button>
     <button class="btn btn-xs btn-warning" ng-click="hide()" translate>{{ content.seeExample }}</button>
   </div>
   <div class="example" ng-hide="messageVisible">
-    {{ content.demoData | translate }} -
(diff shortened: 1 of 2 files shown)


ouranos:
@xaun @cesar-tonnoir See PR. Not sure which branch I should open as it's been an issue for a while.
Happy to rebase on an earlier branch.

I'll need this for KAPPA but I'll create a custom branch with a build include this and #565 for my testing

xaun (approved):
@ouranos Looks like the bug was introduced in 1.9.2, so I guess it's to be merged into 1.9 and tagged as 1.9.3 ?

ouranos:
Hey @xaun, I'm using `v1.8.4` for Kappa and I have the issue so I'd like to have it on `1.8` at least.
If you want me to do it an earlier branch, I'm happy to do so as long as it's then merged in `1.8`.

xaun:
@ouranos @cesar-tonnoir I've looked more closely sorry, and found the bug only exists in 1.8. You can see this by looking at these:

https://github.com/maestrano/impac-angular/blob/1.7/src/services/widgets/widgets.svc.coffee#L223
https://github.com/maestrano/impac-angular/blob/1.8/src/services/widgets/widgets.svc.coffee#L228
https://github.com/maestrano/impac-angular/blob/1.9/src/services/widgets/widgets.svc.coffee#L228

ouranos:
@cesar-tonnoir Could you provide you input on this? Looks like there was an issue with the cascading merged for version?
We had the same issue with the newrelic client version on the backend.

Would be good to review to make sure we don't have other bugs like this that are only on 1.8

cesar-tonnoir (approved):
omg... good old impac-angular v1...
@xaun #goodmemories

LGTM

cesar-tonnoir:
@ouranos I'm not sure what happened there... The last cascading looks correct:

image

I couldn't see any place where we would have merged 1.9 in 1.8?.. :/

ouranos:
Ok, well I'm not sure then, just wanted to point it out. I'll merge this one and we'll keep an eye out.

PR #568, opened 2019-04-18
Fix dashboard creation

ouranos:
Reset the dashboard in scope when opening the modal.
Otherwise the previous dashboard settings are kept.

This fixes the following bug:
- Create a dashboard from a template
- Create a dashboard from scratch
=> The 2nd dashboard is a copy of the template because [dashboard.id is
stil set](https://github.com/maestrano/impac-angular/blob/v1.8.5/src/components/dashboard/dashboard.directive.coffee#L136-L139)


ouranos:
@xaun @cesar-tonnoir Could I get a review on this? This need to be pushed to UAT on Monday.

xaun (approved):
Looks fine to me

cesar-tonnoir (approved):

